PR #28, opened 2023-05-22
feat(cmds): add stacklet commands

Techassi:
This PR adds all service related commands (is actually only one).


nightkr (commented):
Incomplete review, dunno how much is still relevant after today's meeting

src/platform/stacklet/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    CustomCrdNamespaceError { crd_name: String },
+}
+
+/// [`StackletListOptions`] describes available options when listing deployed
+/// services.
+pub struct StackletListOptions {
+    /// Toggle wether to show credentials / secrets in the output. This defaults
+    /// to `false` because of security reasons. Users need to explicitly tell
+    /// the ctl or the web UI to show these credentials.
+    pub show_credentials: bool,
```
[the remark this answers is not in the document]
Techassi: Yes. This is now removed.

src/platform/stacklet/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+            show_versions: true,
+        }
+    }
+}
+
+pub type StackletList = IndexMap<String, Vec<Product>>;
+
+pub fn build_products_gvk_list(
+    product_names: &[&'static str],
+) -> IndexMap<&'static str, GroupVersionKind> {
```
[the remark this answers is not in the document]
Techassi: No, doesn't need to be public. Lifetimes were also adjusted in 7f509b71811ccd5e0e77b803d1e1e80302b0c24f

src/platform/stacklet/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    KubeError { source: KubeError },
+
+    #[snafu(display("no namespace set for custom resource '{crd_name}'"))]
+    CustomCrdNamespaceError { crd_name: String },
+}
+
+/// [`StackletListOptions`] describes available options when listing deployed
+/// services.
+pub struct StackletListOptions {
+    /// Toggle wether to show credentials / secrets in the output. This defaults
```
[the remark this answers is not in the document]
Techassi: Redundant. Was removed.

bins/stackablectl/src/cli/mod.rs
```diff
@@ -215,9 +215,9 @@ pub enum Commands {
     #[command(alias("st"))]
     Stack(StackArgs),
 
-    /// Interact with deployed services of products
+    /// Interact with deployed stacklets
     #[command(alias("svc"))]
```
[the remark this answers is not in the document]
Techassi: I made a clean cut for now. I used the alias `sts`. This feels quite fitting based on that the stack command uses the alias `st`. If we decide the ease migrations, we could also add `svc` and `services` like you suggested.

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -0,0 +1,121 @@
...
+    // namespace.
+    let namespace = args
+        .all_namespaces
+        .then_some(common_args.operator_namespace.as_str());
+
+    let stacklets = list_stacklets(namespace, StackletListOptions::default())
+        .await
+        .context(StackletListSnafu {})?;
+
+    println!("{:?}", stacklets);
```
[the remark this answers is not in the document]
Techassi: Yes, removed.

bins/stackablectl/src/cli/mod.rs
```diff
@@ -215,9 +215,9 @@ pub enum Commands {
     #[command(alias("st"))]
     Stack(StackArgs),
 
-    /// Interact with deployed services of products
+    /// Interact with deployed stacklets
```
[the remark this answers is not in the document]
Techassi: Agreed. Will cook something up.

bins/stackablectl/src/cli/mod.rs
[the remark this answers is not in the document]
nightkr: How about `stl` or `sl`? Not sure about where the last s in `sts` comes from, and it conflicts with K8s' common `STatefulSet` acronym.

bins/stackablectl/src/cli/mod.rs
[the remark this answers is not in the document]
Techassi: Added in f61d058

.github/workflows/build.yml
```diff
@@ -39,7 +40,9 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
         with:
           submodules: recursive
-      - uses: dtolnay/rust-toolchain@1.68.2
+      - uses: dtolnay/rust-toolchain@master
```
nightkr, line 43: These should still be pinned to _something_

Cargo.toml
```diff
@@ -18,24 +18,28 @@ openapi = ["dep:utoipa"]
...
 tera = { workspace = true }
 kube = { workspace = true }
+rand = { workspace = true }
 url = { workspace = true }
 
 [build-dependencies]
 gobuild = "0.1.0-alpha.1"
 
 [workspace.dependencies]
+stackable-operator = { git = "https://github.com/stackabletech/operator-rs.git", rev = "b2ebbc7" }
```
nightkr, line 42: Why does this refer to a commit hash rather than a branch or tag?

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
             }
 
-            Ok(table.to_string())
+            // Only output the error table if there are errors to report.
+            // Currently this is a little awkward, but an upstream PR will make
+            // this more straight forward.
+            Ok(format!(
+                "{}{}",
+                table,
+                if error_table.row_iter().cloned().count() > 0 {
```
nightkr, line 133:
```suggestion
                "{table}{errors}",
                errors = if error_table.row_iter().cloned().count() > 0 {
```

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
+
+/// Processes conditions for a single stacklet / product. It returns a
+/// concatenated string of conditions (which are colored green and red) to
+/// display next to each listed stacklet in the table. Additionally, it also
+/// returns a list of errors to be displayed underneath the stacklet table.
+fn process_conditions(
+    product_conditions: Vec<DisplayCondition>,
+    product_index: &mut usize,
+    use_color: bool,
+) -> (String, Vec<String>) {
```
nightkr, line 153: Might be more clear to return a struct here?

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
                         product.name,
                         product.namespace.unwrap_or_default(),
-                        endpoints,
-                        additional_information,
+                        conditions,
                     ]);
+
+                    for error in errors {
+                        error_table.add_row(vec![product_index.to_string(), error]);
+                    }
```
nightkr, line 123: I'd have to play around with it, but I *think* this formatting will end up being rather confusing.

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
+        OutputType::Json => serde_json::to_string(&stacklets).context(JsonOutputFormatSnafu {}),
+        OutputType::Yaml => serde_yaml::to_string(&stacklets).context(YamlOutputFormatSnafu {}),
+    }
+}
+
+/// Processes conditions for a single stacklet / product. It returns a
+/// concatenated string of conditions (which are colored green and red) to
+/// display next to each listed stacklet in the table. Additionally, it also
+/// returns a list of errors to be displayed underneath the stacklet table.
+fn process_conditions(
```
nightkr, line 149:
```suggestion
fn render_conditions(
```

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
+            use_color,
+        ))
+    }
+
+    (conditions.join(", "), errors)
+}
+
+/// Processes one condition and determines if it is an error (not good). If this
+/// is the case, it get colored red and is returned.
+fn process_condition_error(
```
nightkr, line 181:
```suggestion
fn render_condition_error(
```

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -72,50 +76,146 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
+            )
+        }
+        _ => None,
+    }
+}
+
+/// Colors a single condition (green or red) and additionally adds an error
+/// index to the output.
+fn color_condition(
+    condition: &String,
```
nightkr, line 205:
```suggestion
    condition: &str,
```

src/kube.rs
```diff
@@ -198,10 +197,183 @@ impl KubeClient {
...
+        match product_label {
+            ProductLabel::Both => {
+                params.add_label(format!("app.kubernetes.io/name={product_name}"));
+                params.add_label(format!("app.kubernetes.io/app={product_name}"));
+            }
+            ProductLabel::Name => {
+                params.add_label(format!("app.kubernetes.io/name={product_name}"))
+            }
+            ProductLabel::App => params.add_label(format!("app.kubernetes.io/app={product_name}")),
+        };
```
nightkr:
```suggestion
        if matches!(product_label, ProductLabel::Name | ProductLabel::Both) {
            params.add_label(format!("app.kubernetes.io/name={product_name}"));
        }
        if matches!(product_label, ProductLabel::App | ProductLabel::Both) {
            params.add_label(format!("app.kubernetes.io/app={product_name}"));
        }
```

src/kube.rs
```diff
@@ -198,10 +197,183 @@ impl KubeClient {
...
+    fn add_label(&mut self, label: impl Into<String>) {
+        match self.label_selector.as_mut() {
+            Some(labels) => labels.push_str(format!(",{}", label.into()).as_str()),
+            None => self.label_selector = Some(label.into()),
+        }
+    }
+}
+
+#[derive(Debug, Serialize)]
+pub struct DisplayCondition {
```
nightkr, line 286: This is basically a reified `&dyn ConditionExt`, right?

src/helm.rs
```diff
@@ -26,7 +26,7 @@ pub struct HelmChart {
     pub name: String,
     pub repo: HelmChartRepo,
     pub version: String,
-    pub values: serde_yaml::Value,
+    pub options: serde_yaml::Value,
```
nightkr, line 29: How come?

src/utils/templating.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::constants::PASSWORD_LEN;
+
+pub fn render(content: &str, parameters: &HashMap<String, String>) -> Result<String, tera::Error> {
+    // Create templating context
+    let mut context = Context::new();
+
+    // Fill context with parameters
+    for (name, value) in parameters {
+        context.insert(name, value)
+    }
```
nightkr, line 16:
```suggestion
    let mut context = Context::from_serialize(parameters)?;
```

src/utils/templating.rs
```diff
@@ -0,0 +1,52 @@
...
+    let mut tera = Tera::default();
+    tera.register_function("random_password", random_password());
+    tera.register_function("bcrypt", bcrypt());
+
+    // Render template
+    tera.render_str(content, &context)
+}
+
+fn random_password() -> impl Function {
+    Box::new(
```
nightkr, line 28: No need to box this, as far as I can tell from the signatures?

src/utils/templating.rs
```diff
@@ -0,0 +1,52 @@
...
+        move |_args: &HashMap<String, Value>| -> tera::Result<Value> {
+            let password = Alphanumeric.sample_string(&mut rand::thread_rng(), PASSWORD_LEN);
+            Ok(password.into())
+        },
+    )
+}
+
+fn bcrypt() -> impl Function {
+    Box::new(
```
nightkr, line 37: Same here.

src/helm.rs
Techassi: This is needed because otherwise the deserialization fails. The templated stacks use `options`, e.g. [here](https://github.com/stackabletech/stackablectl/blob/main/stacks/_templates/jupyterhub.yaml#L7).

nightkr: Well that sucks, but I can see why we're kind of stuck with it now...

src/utils/templating.rs
Techassi: I followed the official docs [here](https://tera.netlify.app/docs/#functions).

nightkr: Odd, [the API docs](https://docs.rs/tera/latest/tera/trait.Function.html#implementors) only mention a generic impl over `Fn`.

.github/workflows/build.yml
Techassi: The recommended way of doing this is described [here](https://github.com/dtolnay/rust-toolchain#inputs). The `@rev` is used to select the Rust toolchain (not the action revision). This change is needed to use a Rust version defined by a variable (See `RUST_VERSION` above).

src/utils/templating.rs
nightkr: (Which should still work with the boxing, but only because `Box<T>: Fn where T: Fn`.

Cargo.toml
Techassi: Changed in 939d5e3

.github/workflows/build.yml
nightkr: No, we should be pinning both the action and the Rust toolchain separately. Something like `uses: dtolnay/rust-toolchain@0e66bd3e6b38ec0ad5312288c83e47c143e6b09e # v1`. The old pin was also wrong, since the Rust versions revs are floating branches and updated whenever master is (from the looks of it, anyway).

src/helm.rs
Techassi: Agreed. We would have to change that in the old `stackablectl` and the stack files. We can do that once we migrate the the new one and the improved demo/stack/release repo structure.

src/utils/templating.rs
Techassi: Odd indeed. I now removed the `Box` in f33a89f

.github/workflows/build.yml
Techassi: I obviously agree that we should pin both version. I'm not quite sure if pinning the action will work tho (as the `@rev` is parsed as the toolchain version). I will test that.

Techassi: Pinned in 914e1a6

bins/stackablectl/src/cli/mod.rs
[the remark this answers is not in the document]
Techassi: Oh shoot, you are right. Will remove the current alias and add your suggested ones.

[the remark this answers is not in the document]
Techassi: Changed in ec01086

src/utils/templating.rs
Techassi: Changed in aae43f5

src/kube.rs
Techassi: Changed in aae43f5

bins/stackablectl/src/cmds/stacklets.rs
Techassi: Improved in 0153ed2

bins/stackablectl/src/cmds/stacklets.rs
Techassi: Agreed, will update.

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -142,80 +133,81 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
-                    .to_string(),
-            )
-        }
+        (Some(false), true) => Some(
+            Red.paint(format!("[{}]: {}", error_index, message))
+                .to_string(),
+        ),
         _ => None,
     }
 }
```
nightkr: Hold on, doesn't this mean that we don't render the message at all if colour is disabled? How about:

```rust
if !is_good.unwrap_or(true) {
    *condition_index += 1;
    let mut message = format!("[{condition_index}]: {message}");
    if use_color {
        message = Red.paint(message).to_string();
    }
    Some(message)
} else {
    None
}
```

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -142,80 +133,81 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
     use_color: bool,
 ) -> String {
     match (is_good, use_color) {
         (Some(true), true) => Green.paint(condition).to_string(),
         (Some(false), true) => Red
-            .paint(format!(
-                "{}: See [{}.{}]",
-                condition, product_index, condition_index
-            ))
+            .paint(format!("{}: See [{}]", condition, error_index))
```
nightkr, line 198:
```suggestion
            .paint(format!("{condition}: See [{error_index}]"))
```

bins/stackablectl/src/cmds/stacklets.rs
```diff
@@ -142,80 +133,81 @@ async fn list_cmd(args: &StackletListArgs, common_args: &Cli) -> Result<String,
...
             .to_string(),
         _ => condition.to_owned(),
```
nightkr, line 200: Don't we still want to render the reference (if required) even without colour?

bins/stackablectl/src/cmds/stacklets.rs
Techassi: Changed in 6b1d4ff

bins/stackablectl/src/cmds/stacklets.rs
Techassi: You are absolutely correct, fixed in bf5960b

bins/stackablectl/src/cmds/stacklets.rs
Techassi: Changed in bf5960b

bins/stackablectl/src/cmds/stacklets.rs
Techassi: Yes! Updated in bf5960b

nightkr (approved):

PR #32, opened 2023-05-25
Add TypeScript helper types

Techassi:
This PR adds the `Result` and `Option` helper types. They are closely modeled after the Rust types.


web/src/types/option.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export const Some = <S extends ToString>(value: S) => {
+  return new Option<S>(false, value);
+};
+
+export const None = () => {
+  return new Option<never>(true);
+};
+
+export class Option<S extends ToString> {
```
nightkr, line 13: Do we really need this `ToString` bound here?

web/src/types/option.ts
```diff
@@ -0,0 +1,199 @@
...
+  private is_none: boolean;
```
nightkr, line 14: I don't know, I think I'd prefer to use subtypes for this rather than boolean flags, but I'm guessing there's a reason it's done this way?

web/src/types/option.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (this.isSome() && other.isNone()) {
+      return this;
+    } else if (this.isNone() && other.isSome()) {
+      return other;
+    } else {
+      return None();
+    }
+  }
+
+  take(): Option<S> {
```
nightkr, line 124: `take()` isn't really useful in a language without borrow checking IMO, and it's kind of surprising for a type that I'd otherwise expect to be immutable.

web/src/types/result.ts
```diff
@@ -0,0 +1,136 @@
...
+      ? matcher.err(this.value as E)
+      : matcher.ok(this.value as O);
+  }
+
+  unwrap(): O {
+    return this.expect('called `Result.unwrap()` on an `Err` value');
+  }
+
+  unwrapErr(): E {
+    return this.expectErr('called Result.unwrapErr() on an `Ok` value');
```
nightkr, line 57:
```suggestion
    return this.expectErr('called `Result.unwrapErr()` on an `Ok` value');
```

web/src/types/option.ts
```diff
@@ -0,0 +1,199 @@
...
+  return new Option<S>(false, value);
+};
+
+export const None = () => {
+  return new Option<never>(true);
+};
+
+export class Option<S extends ToString> {
+  private is_none: boolean;
+  private value: S | never | undefined;
```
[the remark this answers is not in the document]
Techassi: In this case: no. The option either had a value of type `S` or when `None` of type `never`. This was needed, as the Option was modeled as one type and not two distinct `Some` and `None` types. This was changed in bad24a5

web/src/types/option.ts
Techassi: It was needed for the `okOr` and `okOrElse` methods, because the `Ok` value was part of the error message when calling `unwrapErr`. I removed the value from the error message (in bad24a5) and thus we don't need this bound anymore.

web/src/types/option.ts
Techassi: Option now uses subtypes. Changed in bad24a5

web/src/types/option.ts
Techassi: Agreed, removed in bad24a5

Techassi:
I re-modeled `Option` to use subtypes. Should we re-model `Result` as well? I would say no, as it works pretty good in the current from.

nightkr:
I'd say yes, since we're making a bunch of assertions around the type of `Result.value`...

nightkr:
Proposed a few more simplifications at https://github.com/stackabletech/stackable-cockpit/compare/feature/web-helper-types...feature/web-helper-types-/nat-option-simplify

Techassi:
I merged your proposed changes. Will rework the `Result` type as well in the following days.

Techassi:
Okay after looking it over again, I think the result implementation is fine as is. It kinda bloats the implementation when we re-model this with subtypes. The `Result` type doesn't suffer from the weirdness like `Option` (having no value with the `None` variant). Instead we can perfectly model both the `Ok` and the `Err` value with one type.

I would suggest we move this PR forward as is.

nightkr:
It has the same weirdness, you can't get out of the casting hell in at least `match()` (and implicitly in `isOk`/`isErr`). But oh well.

nightkr:
In theory you could at least reduce it to `isOk`/`isErr`, but I don't think TypeScript lets us model that we're _always_ `isOk XOR isErr`.

Techassi:
I re-modeled the Result type as well. It now also uses subtypes.

web/src/types/result.ts
```diff
@@ -1,136 +1,330 @@
...
-  }): T | K {
-    return this.is_error
-      ? matcher.err(this.value as E)
-      : matcher.ok(this.value as O);
+  match<T>(matcher: { ok: (value: Readonly<O>) => T }): T {
+    return matcher.ok(this.value);
   }
 
   unwrap(): O {
     return this.expect('called `Result.unwrap()` on an `Err` value');
```
nightkr, line 198:
```suggestion
    return this.value;
```

web/src/types/result.ts
```diff
@@ -1,136 +1,330 @@
...
+  match<T, K>(matcher: { err: (error: Readonly<E>) => K }): T | K {
+    return matcher.err(this.error);
+  }
+
+  unwrap(): never {
+    return this.expect('called `Result.unwrap()` on an `Err` value');
+  }
+
+  unwrapErr(): E {
+    return this.expectErr('called `Result.unwrapErr()` on an `Ok` value');
```
nightkr, line 280:
```suggestion
    return this.error;
```

nightkr (approved):

Issue #14, opened 2023-05-15
Show list of stacklets (ZooKeepers, HDFS clusters, etc)


nightkr:
```[tasklist]
### Tasks
- [x] Show basic table
- [ ] Connect table to K8s API
- [ ] Show stacklet status
```

lfrancke:
This is in Done and closed but only one of the items has been checked.
Can you update the ticket?

lfrancke:
@nightkr ping

nightkr:
Sorry, yeah the closing PR should have handled both of those. Updating now.

PR #34, opened 2023-05-25
[Merged by Bors] - Start connecting the UI to the API

### LINKED ISSUE #33: Connect UI to API

nightkr:
# Description

This loads the list of stacklets from a mock endpoint in the API, with a type schema generated by utoipa.

It also fixes up the cases where OpenAPI didn't quite line up with the actual APIs.

Fixes #33, part of #14.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


.pre-commit-config.yaml
```diff
@@ -50,8 +52,16 @@ repos:
...
         language: system
         entry: cargo xtask gen-comp
         stages: [commit, merge-commit, manual]
+        pass_filenames: false
+      - id: gen-openapi
+        name: gen-openapi
+        language: system
+        entry: cargo xtask gen-openapi
+        stages: [commit, merge-commit, manual]
+        pass_filenames: false
```
nightkr, line 67: Ideally this would be a vite plugin rather than a pre-commit hook, but... vite's plugin API is honestly quite bewildering to me at the moment.

bins/stackabled/src/handlers/stacklets.rs
```diff
@@ -0,0 +1,41 @@
+use axum::{extract::Path, routing::get, Json, Router};
+use stackable::kube::k8s_openapi::{
+    apimachinery::pkg::apis::meta::v1::ObjectMeta, serde::Serialize,
```
Techassi, line 3: Do we want to really want to use the re-exported `k8s_openapi` and `serde` crates. I favor direct imports and dependencies in the `Cargo.toml` file.

bins/stackabled/src/handlers/utoipa_synthetic.rs
Techassi, line 1: I think we should place these "synthetic" types in their own module, like `types` or `synthetic`.

xtask/src/main.rs
```diff
@@ -1,55 +1,91 @@
...
-use std::{env, fs};
+use std::{
+    fs,
+    io::Write,
+    process::{Command, Stdio},
+};
+
+#[derive(clap::Parser)]
+#[allow(clippy::enum_variant_names)]
+enum XtaskCli {
```
Techassi, line 16:
```suggestion
enum XtaskCommand {
```

This name fits better for what xtask actually does.

web/src/pages/stacklets/list.tsx
```diff
@@ -1,26 +1,26 @@
...
+      <Show when={stacklets.loading}>Loading...</Show>
       <DataTable
-        items={productClusters() || []}
+        items={stacklets() || []}
         columns={[
           {
             label: 'Product',
-            get: (cluster) => cluster.product,
-            sortBy: (cluster) => cluster.product,
+            get: (x) => x.product,
```
Techassi, line 17: I feel like `x` is kinda non-descriptive. Why don't we use `stacklet` or `stack`?

.pre-commit-config.yaml
Techassi: Agreed. This is part of the build process and not necessarily a pre-commit "check".

Techassi (changes requested):
Mostly minor stuff. Rest looks good!

bins/stackabled/src/handlers/stacklets.rs
nightkr: Happy to add a direct dependency on `serde`. `k8s-openapi` is a bit weird since we need to have one canonical place that decides which version features to use.

xtask/src/main.rs
nightkr: Works for me, I was just trying to be consistent with it being called `Cli` everywhere else.

bins/stackabled/src/handlers/stacklets.rs
nightkr: Then again, this is more of a temp solution anyway, `Stacklet` should be moved to libcockpit at some point.

nightkr: ba8c284cd337680db604d546438bd3db84ebbeec

bins/stackabled/src/handlers/utoipa_synthetic.rs
nightkr: c09e62adeaa12bf19d2005983d808dda7c2baf04

nightkr: I think it would make sense to do this for all handlers, actually...

nightkr: 8631f3b663f25e00c9d48e2ff63b14dd1536001d, 856022e39da40aa6e8c971c27210eb25b0f5a0b5

xtask/src/main.rs
nightkr: 43a044ae3e9b6f9a05ea3ec703ebf5c05de9c3e5

xtask/src/main.rs
```diff
@@ -1,55 +1,91 @@
...
+                .take()
+                .expect("child stdin must be available")
+                .write_all(openapi_json.as_bytes())
+                .context(WriteOpenapiSchemaSnafu)?;
+            let status = codegen.wait().context(ImportOpenapiSchemaRunSnafu)?;
+            if !status.success() {
+                return ImportOpenapiSchemaSnafu {
+                    error_code: status.code(),
+                }
+                .fail();
```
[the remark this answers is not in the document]
nightkr: That would just cause the error to be constructed and immediately dropped.

web/src/pages/listeners.tsx
```diff
@@ -1,5 +1,5 @@
 import { Show, For, Switch, Match, createResource } from 'solid-js';
-import { getListeners } from '../api';
+import { getListeners } from '../api/index';
```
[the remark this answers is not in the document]
nightkr: I can't remember where exactly, but I got errors _somewhere_ when trying to import the folder directly. I think I remember running into it having something to do with Node supporting it as legacy behaviour.

https://github.com/vitejs/vite/issues/6193 seems related, but I'm not sure if it was the exact source I found back then.

.pre-commit-config.yaml
nightkr: There's also a bit of a circular dependency problem.. to avoid that we'd need to split `cockpit-api` into a separate crate that knows how to emit its API, which we then combine with the `cockpit-ui` to build the `cockpitd` server binary that people run.

web/src/pages/stacklets/list.tsx
nightkr: Mostly, just that it's a lot of extra typing and visual noise when we already know from context (the `items` parameter) that we're working with a list of stacklets.

I'd consider it similar to `stacklets.map((x) => x.metadata.name)` vs `stacklets.map((stacklet) => stacklet.metadata.name)`.

I'm specifically not using `stack`, since we've talked before about using that to describe a whole use-case deployment (think: roughly the same scope as our demos).

.pre-commit-config.yaml
nightkr: Happy to look at this at some point, but I think it deserves its own tech-debt ticket rather than blocking this one.

bins/stackabled/src/handlers/utoipa_synthetic.rs
Techassi: Yes, that's great!

bins/stackabled/src/handlers/stacklets.rs
Techassi: > Happy to add a direct dependency on `serde`. `k8s-openapi` is a bit weird since we need to have one canonical place that decides which version features to use.

We can "pin" the version feature used across all workspace members using a workspace dependency. With that, all members using that dependency, use the same version.

web/src/pages/listeners.tsx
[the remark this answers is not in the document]
Techassi: Mhm this seems weird. Let's see if this issue keeps coming up or if we can drop the suffix again.

.pre-commit-config.yaml
Techassi: Yes, would be great if you can look into this. But I agree, this shouldn't be a blocker for this PR.

bins/stackabled/src/handlers/stacklets.rs
nightkr: True, good catch.

nightkr: b00fbbe350fd99562fb6f64c09166271934d7076

xtask/src/main.rs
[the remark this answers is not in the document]
nightkr: 0262b7724cda5f7423f37b1656450732ae165405

web/src/pages/listeners.tsx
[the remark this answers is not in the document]
nightkr: I can't seem to reproduce the issue now, reverted it in 0e294c452a50142c68eea56c78fe0638791f37a9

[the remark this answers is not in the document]
Techassi: Okay great! This is resolved then.

xtask/src/main.rs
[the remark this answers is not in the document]
Techassi: Looks good!

bins/stackabled/src/handlers/stacklets.rs
Techassi: Neat! Resolved.

web/src/pages/stacklets/list.tsx
Techassi: Yeah I agree that `stack` is not the right name in this context at all! Let's stick with `x` for now as the code can get a little cluttered with longer names (as you mentioned). If we decide we want something other than `x`, this will be a simple find and replace.

bins/stackabled/Cargo.toml
```diff
@@ -18,3 +18,5 @@ tracing = { workspace = true }
 snafu = { workspace = true }
 tokio = { workspace = true }
 stackabled-web = { path = "../../web", optional = true }
+serde.workspace = true
+k8s-openapi.workspace = true
```
Techassi, line 22: Maybe we should be consistent which format we use. But this is a nitpick.

```
serde = { workspace = true }
# VS
serde.workspace = true
```

xtask/Cargo.toml
```diff
@@ -5,7 +5,9 @@ edition = "2021"
 
 [dependencies]
 stackablectl = { path = "../bins/stackablectl" }
+stackabled = { path = "../bins/stackabled" }
 clap_complete =  { workspace = true }
 snafu = { workspace = true }
-clap = { workspace = true }
+clap.workspace = true
 clap_mangen = "0.2.10"
+serde_json.workspace = true
```
Techassi, line 13: Same as above.

Techassi (approved):
Looks good!

bins/stackabled/Cargo.toml
nightkr: The latter is what `cargo add` adds if the dependency is already listed in `workspace.dependencies`, so I'd suggest standardizing on that.

PR #37, opened 2023-06-01
[Merged by Bors] - Update comfy-table

Techassi:
Updates `comfy-table` and makes use of the new `add_row_if` method.


sbernauer (approved):
Thx for upstream contribution ;)

Issue #40, opened 2023-06-07
Tracking: feature parity / new features for stackablectl


Techassi:
This issue tracks the feature parity of the new stckablectl compared to the current one.

```[tasklist]
### Pull Requests for Feature Parity
- [ ] https://github.com/stackabletech/stackable-cockpit/pull/28
- [ ] Build for all the supported platforms (we might need to vendor openssl)
```

```[tasklist]
### New features
- [ ] https://github.com/stackabletech/stackable-cockpit/pull/35
- [ ] https://github.com/stackabletech/stackable-cockpit/issues/39
- [ ] Add support to uninstall demo
```
```[tasklist]
### Improvements
- [ ] https://github.com/stackabletech/stackable-cockpit/issues/41
```

PR #47, opened 2023-06-22
feat!: add transfer and caching layer

### LINKED ISSUE #39: Add transfer / caching layer
Add a caching layer for every network call there is in `stackablectl`. Currently only a few files (the demo, stack and release files) use the caching mechanism.

Adding this caching layer for every network call requires a central network request mechanism through which every request is funneled. This can be achieved by a central struct, which wraps a `reqwest` client. Alongside this client, the caching layer handles retrieving and storing cache files.

Techassi:
Fixes #39 

This PR adds a central transfer and caching layer for all network requests for remote resources like stack, release and demo lists.


src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn extract_file_name(url: &Url) -> Result<String> {
+        let segment = url
+            .path_segments()
+            .ok_or(TransferError::FileNameError)?
+            .last()
+            .ok_or(TransferError::FileNameError)?;
+
+        ensure!(segment.contains('.'), FileNameSnafu {});
+        Ok(segment.to_string())
+    }
```
nightkr, line 143: Why isn't this in the cache layer instead? It's only used for that, right?

src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Returns the last URL path segment as the filename. It ensures that the
+    /// last segment contains at least one dot.
+    fn extract_file_name(url: &Url) -> Result<String> {
+        let segment = url
+            .path_segments()
+            .ok_or(TransferError::FileNameError)?
+            .last()
+            .ok_or(TransferError::FileNameError)?;
```
nightkr, line 139: This is susceptible to a lot of cache confusion attacks (`https://stackable.tech/stacks.yaml` would get the same cache key as `https://evil.com/stacks.yaml`). Instead, I'd go with something like `{sanitized_path}-{sha256}` (where `sanitized_path = url.as_str().replace(|chr| !chr.is_ascii_alphanumeric(), '-')`).

src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
+use std::collections::HashMap;
+
+use serde::Deserialize;
+use snafu::{ensure, ResultExt, Snafu};
+use url::Url;
+
+mod cache;
+pub use cache::*;
```
nightkr, line 8: Does this need to be `pub`? If so, why not make the module itself public instead of reexporting it?

src/xfer/mod.rs
Techassi: Yes. I'm aware of it. I was debating if it is worth for use, as we only deal with known URLs. But... I can add this feature!

src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        url: &Url,
+        parameters: &HashMap<String, String>,
+    ) -> Result<T>
+    where
+        T: for<'a> Deserialize<'a> + Sized,
+    {
+        let content = self.get_from_cache_or_remote(url).await?;
+        let content = templating::render(&content, parameters).context(TemplatingSnafu)?;
+        serde_yaml::from_str(&content).context(YamlSnafu {})
+    }
```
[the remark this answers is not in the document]
nightkr: One alternate way to handle it: https://github.com/stackabletech/stackable-cockpit/compare/feature/xfer-layer...feature/xfer-layer-/parsers

src/xfer/mod.rs
nightkr: I would have assumed that anything that can take a local path (like `List::build`) could also take an arbitrary URL?

Either way, this is as much about avoiding confusing bugs in the future as it is about _security_ per se.

src/platform/stack/spec.rs
```diff
@@ -160,11 +172,18 @@ impl StackSpecV2 {
                     debug!("Installing manifest from Helm chart {}", helm_file);
 
                     // Read Helm chart YAML and apply templating
-                    let helm_chart = read_yaml_data_with_templating(helm_file, parameters)
-                        .await
-                        .context(TemplatedReadSnafu)?;
                     let helm_chart: HelmChart =
-                        serde_yaml::from_str(&helm_chart).context(YamlSnafu)?;
+                        match helm_file.into_path_or_url().context(PathOrUrlParseSnafu)? {
```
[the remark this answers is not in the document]
Techassi: I was actually thinking about combining local file AND remote file handling into the same `Client`. However, I wasn't sure if it is the right call. But I can include the local file handling and the `Client` then consumes a `PathOrUrl` instead of just `Url`.

src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            .get(url.clone())
+            .build()
+            .context(RequestSnafu {})?;
+        let result = self.client.execute(req).await.context(RequestSnafu {})?;
+
+        result.text().await.context(RequestSnafu {})
+    }
+
+    /// Returns the last URL path segment as the filename. It ensures that the
+    /// last segment contains at least one dot.
```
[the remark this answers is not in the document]
Techassi: I'm also on the edge about this. This is a "check" (if you can even call it that) that the URL points to a file "with" an extension. We can drop this however when we introduce the hashed filenames (which have no extensions).

src/xfer/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    TemplatingError { source: tera::Error },
+}
+
+#[derive(Debug)]
+pub struct TransferClient {
+    pub(crate) client: reqwest::Client,
+    pub(crate) cache: Cache,
+}
+
+impl TransferClient {
```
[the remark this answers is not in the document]
Techassi: Renamed in 79c0498

src/xfer/mod.rs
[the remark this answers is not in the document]
Techassi: Very clean solution, I like it. Implemented a slight variation in 79da379

src/platform/stack/spec.rs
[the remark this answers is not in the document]
Techassi: The client now takes a parameter of type `PathOrUrl` since 79da379

src/xfer/mod.rs
Techassi: Removed in 37e7c2d

package.json
```diff
@@ -2,5 +2,6 @@
   "private": true,
   "workspaces": [
     "web"
-  ]
+  ],
+  "packageManager": "yarn@1.22.19"
```
sbernauer, line 6: No clue if this intentional ;)

sbernauer (approved):
LGTM, but might makes sense to wait for @nightkr to review

sbernauer:
One small thing: Maybe we want to decrease the cache duration from 1h to 15min or so

web/package.json
```diff
@@ -67,5 +67,6 @@
         ]
       }
     }
-  }
+  },
+  "packageManager": "yarn@1.22.19"
```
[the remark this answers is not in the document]
Techassi: It is :)

src/xfer/cache.rs
```diff
@@ -0,0 +1,210 @@
...
+#[derive(Debug)]
+pub struct Cache {
+    pub(crate) settings: CacheSettings,
+}
+
+impl Cache {
+    /// Creates a new [`Cache`] instance with the provided `settings`. It should
+    /// be noted that it is required to call the [`Cache::init`] method before
+    /// using the cache for the first time to ensure the backend is setup
+    /// properly.
```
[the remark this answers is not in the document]
Techassi: The type state pattern is actually a nice idea!

bins/stackablectl/src/cmds/demo.rs
```diff
@@ -124,26 +119,33 @@ pub enum DemoCmdError {
...
             .context(ListSnafu)?;
 
         match &self.subcommand {
             DemoCommands::List(args) => list_cmd(args, list).await,
             DemoCommands::Describe(args) => describe_cmd(args, list).await,
-            DemoCommands::Install(args) => install_cmd(args, common_args, list).await,
-            // DemoCommands::Uninstall(args) => uninstall_cmd(args, list),
+            DemoCommands::Install(args) => {
+                install_cmd(args, common_args, list, &transfer_client).await
+            } // DemoCommands::Uninstall(args) => uninstall_cmd(args, list),
```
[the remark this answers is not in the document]
Techassi: Yeah I just commented it out as we currently have no way of doing this. I can also remove it completely. We should add it to our TODO list tho.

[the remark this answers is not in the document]
Techassi: Added, see #40

Techassi:
Yeah I wasn't quite sure how long the default cache time should be. What are you reasons to only make it 15 minutes instead of one hour?

bins/stackablectl/src/cmds/demo.rs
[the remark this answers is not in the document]
Techassi: Removed code in 9776739

src/xfer/cache.rs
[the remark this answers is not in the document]
Techassi: Added in e126f60

package.json
nightkr: It is, yeah.

src/xfer/cache.rs
```diff
@@ -25,38 +32,56 @@ pub enum CacheError {
 }
 
 #[derive(Debug)]
-pub struct Cache {
+pub struct Cache<State> {
     pub(crate) settings: CacheSettings,
+    state: PhantomData<State>,
```
nightkr, line 37: Both states are ZSTs anyway, so there's no need to wrap them in `PhantomData`.

src/xfer/cache.rs
```diff
@@ -25,38 +32,56 @@ pub enum CacheError {
...
+}
+
+impl<State> Cache<State> {
```
nightkr, line 40: Is there a reason you didn't do the more typical `CacheBuilder`/`Cache` split here?

nightkr: Actually, isn't the `CacheSettings` effectively the builder in this case anyway? Where would you prefer `Cache<CacheUninitialized>` over `CacheSettings`?

src/xfer/cache.rs
```diff
@@ -0,0 +1,235 @@
...
+pub type CacheResult<T> = Result<T, CacheError>;
+
+#[derive(Debug)]
+pub struct CacheUninitialized;
+
+#[derive(Debug)]
+pub struct CacheInitialized;
+
+#[derive(Debug, Snafu)]
+pub enum CacheError {
```
nightkr, line 23:
```suggestion
pub enum Error {
```

No need to prefix everything with the module name.

src/xfer/cache.rs
```diff
@@ -0,0 +1,235 @@
...
+
+use crate::constants::DEFAULT_CACHE_MAX_AGE;
+
+pub type CacheResult<T> = Result<T, CacheError>;
+
+#[derive(Debug)]
+pub struct CacheUninitialized;
+
+#[derive(Debug)]
+pub struct CacheInitialized;
```
nightkr, line 20: Same here.

src/xfer/cache.rs
```diff
@@ -0,0 +1,235 @@
...
+};
+
+use sha2::{Digest, Sha256};
+use snafu::{ResultExt, Snafu};
+use tokio::{fs, io};
+use url::Url;
+
+use crate::constants::DEFAULT_CACHE_MAX_AGE;
+
+pub type CacheResult<T> = Result<T, CacheError>;
```
nightkr, line 14:
```suggestion
type Result<T, E = Error> = Result<T, E>;
```

src/xfer/cache.rs
Techassi: I think otherwise the struct complains about `State` not being used. You can avoid using `PhantomData` when the struct uses the generic type (which we don't need / do here).

src/xfer/cache.rs
Techassi: d187c02

src/xfer/cache.rs
Techassi: d187c02

src/xfer/cache.rs
nightkr: Renaming the field to begin with `_` should also silence that warning (`_state`).

bins/stackablectl/src/cmds/cache.rs
```diff
@@ -26,39 +26,29 @@ pub enum CacheCommands {
...
 pub enum CacheCmdError {
-    #[snafu(display("io error: {source}"))]
-    IoError { source: io::Error },
+    #[snafu(display("cache settings error"))]
+    CacheSettingsError { source: CacheSettingsError },
 
-    #[snafu(display("xdg error: {source}"))]
-    XdgError { source: BaseDirectoriesError },
+    #[snafu(display("cache error"))]
+    CacheError { source: Error },
```
nightkr, line 33:
```suggestion
    CacheError { source: cache::Error },
```

Techassi: d2c001d

src/xfer/cache.rs
Techassi: 1d5afed

src/xfer/cache.rs
```diff
@@ -31,65 +24,77 @@ pub enum Error {
     WriteDisabled,
 }
 
-#[derive(Debug)]
-pub struct Cache<State> {
-    pub(crate) settings: CacheSettings,
-    state: PhantomData<State>,
+pub struct CacheBuilder {
+    backend: CacheBackend,
+    max_age: Duration,
```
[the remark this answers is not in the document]
Techassi: Kinda true. I would favor `CacheSettings::into_cache(self)` which handles the (fallible, but required) initialization. `try_into_cache` might be a better name then.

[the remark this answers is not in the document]
Techassi: Or we remove `CacheSettings` all together...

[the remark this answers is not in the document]
Techassi: Added `try_into_cache` in 2d26b59

nightkr (approved):

PR #48, opened 2023-06-28
Basic authentication

### LINKED ISSUE #17: Implement MVP (config-bound single user) authentication

nightkr:
# Description

Fixes #17

The API design here is by no means final, we just want to present _something_.

A configured username/password is resolved into a session token (so that we can keep the user logged in with _somewhat_ good conscience). Session tokens don't currently expire or persist beyond a server restart, both of those should be broken into separate issues for later. We should also consider alternate login methods (such as OAuth).

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


bins/stackabled/src/cli.rs
```diff
@@ -16,4 +16,10 @@ pub struct Cli {
     /// Address the server binds to
     #[arg(short, long, default_value = "127.0.0.1", env = "STACKABLED_ADDRESS")]
     pub address: IpAddr,
+
+    /// Path to the password database, can be generated with the Apache `htpasswd` utility
+    ///
+    /// Only bcrypt passwords are supported (`htpasswd -B`).
+    #[arg(long, env = "STACKABLED_HTPASSWD")]
```
Techassi, line 23: `STACKABLED_HTPASSWD_PATH` or `STACKABLED_HTPASSWD_FILE` might be better here.

bins/stackabled/src/middleware/authentication.rs
```diff
@@ -0,0 +1,173 @@
...
+                        PasswordHash::Bcrypt(prefixed_pw_hash.to_string()),
+                    );
+                    Ok(())
+                } else {
+                    InvalidHashTypeSnafu.fail()
+                }
+            } else {
+                NoSeparatorSnafu.fail()
+            }
+            .context(EntrySnafu { line })?
```
Techassi, line 96: This implementation looks solid, but is there a crate out there to pass these files? If not, we can keeps it. It look simple enough.

bins/stackabled/src/middleware/authentication.rs
```diff
@@ -0,0 +1,173 @@
...
+            let password_hash = self.state.accounts.get(username);
+            if password_hash
+                .and_then(|PasswordHash::Bcrypt(hash)| {
+                    bcrypt::verify(basic_creds.password(), hash).ok()
+                })
+                .unwrap_or(false)
+            {
+                Ok(Username(username.to_string()))
+            } else {
+                Err("request authentication failed: invalid username or password")
```
Techassi, line 131: Might be worth to type those errors.

bins/stackabled/src/middleware/authentication.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        } else {
+            Err("request authentication failed: no valid authentication method presented")
```
Techassi, line 134: Should we include the supported methods in the error message?

Techassi (commented):
Partial review: I looked at the Rust code. Will follow-up with the TypeScript code asap.

bins/stackabled/src/middleware/authentication.rs
nightkr: Ah yes, the eternal lure of "I'll get to that, *eventually*". Always promptly followed by forgetting about it completely.

bins/stackabled/src/middleware/authentication.rs
Techassi: That sounds reasonable. It might be worth to pull this code out into its own file and maybe add some tests. In the future we might even put it in a separate crate.

bins/stackabled/src/middleware/authentication.rs
Techassi: Happens :)

web/src/api/index.ts
```diff
@@ -1,7 +1,37 @@
...
+          setCurrentSessionToken();
+        }
+      })
+      // We don't want to block page loads on waiting for this validation
+      // eslint-disable-next-line unicorn/prefer-top-level-await
+      .catch((error) =>
+        console.error('Failed to validate session token', error),
+      );
+  }
+}
```
Techassi, line 34: I'm not a super huge fan of this being executed automagically. We might want to move this to an `initAuth` function, which gets called when the page loads.

web/src/api/index.ts
```diff
@@ -1,7 +1,37 @@
 import createClient from 'openapi-fetch';
 import { components, paths } from './schema';
+import { createLocalStorageSignal } from '../utils/localstorage';
 
 const client = createClient<paths>({ baseUrl: '/api' });
+const [currentSessionToken, setCurrentSessionToken] =
```
Techassi, line 6: It might be worth to start splitting different API calls into different files, e.g. `api/auth` and `api/stacklets`

web/src/api/index.ts
```diff
@@ -23,6 +53,24 @@ function delay(amount: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, amount));
 }
 
+export async function logIn(
+  username: string,
+  password: string,
+): Promise<string | undefined> {
+  const response = await client.post('/login', {
+    headers: { Authorization: 'Basic ' + btoa(username + ':' + password) },
```
Techassi, line 61: Use template strings, aka backticks:

```ts
headers: { Authorization: `Basic ${btoa(`${username}:${password}`}` }
```

web/src/api/index.ts
```diff
@@ -23,6 +53,24 @@ function delay(amount: number): Promise<void> {
...
+  });
+  setCurrentSessionToken(response.data);
+  if (!response.response.ok) {
+    return response.error;
```
Techassi, line 65: This is fine for now, but we should have a discussion about error handling in the frontend and how we display these to the user.

web/src/api/index.ts
```diff
@@ -82,7 +130,7 @@ export async function getListeners(): Promise<Listener[]> {
 
 type Stacklet = components['schemas']['Stacklet'];
 export async function getStacklets(): Promise<Stacklet[]> {
-  const { data } = await client.get('/stacklets', {});
+  const { data } = await client.get('/stacklets', sessionOptions());
```
Techassi, line 133: I guess we should pass all required options (like the auth header) when creating the client. This ensures that every API call is done with the required header. This however means to things:

- We might need a separate client which handles auth (without the header included)
- We have to wrap the client in our own logic to provide dynamic session token updating via signals

web/src/pages/login.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <>
+      <Show when={isLoggedIn()}>{props.children}</Show>
+      <Show when={!isLoggedIn()}>
+        <LoginPage />
+      </Show>
+    </>
+  );
+};
+
+export const LoginPage = () => {
```
Techassi, line 18: Here might be a good place to place the above mentioned `initAuth` function.

web/src/pages/login.tsx
```diff
@@ -0,0 +1,61 @@
+import { JSX, Show, createResource, createSignal } from 'solid-js';
+import { isLoggedIn, logIn } from '../api';
+
+interface LoginPageOrProps {
+  children: JSX.Element;
+}
+export const LoginPageOr = (props: LoginPageOrProps) => {
+  return (
+    <>
+      <Show when={isLoggedIn()}>{props.children}</Show>
```
Techassi, line 10: We can get away with only one `<Show />` when we use the `fallback` prop.

web/src/utils/localstorage.ts
```diff
@@ -0,0 +1,28 @@
...
+function createStorageSignal(
+  storage: Storage,
+  name: string,
+): Signal<string | undefined> {
+  const [value, setValue] = createSignal<string | undefined>(
+    storage.getItem(name) || undefined,
+  );
+  // The session writer effect's lifecycle should be tied to that of the signal
+  createRoot(() => {
+    createEffect(() => {
```
Techassi, line 12: Should we re-validate this value after every render call? Might be a little excessive.

Techassi (commented):
TypeScript code done

web/src/api/index.ts
nightkr: Yep definitely.

bins/stackabled/src/middleware/authentication.rs
nightkr: 741c431842a533579073f134e66ab3dfa7b900e4

bins/stackabled/src/cli.rs
nightkr: 15f60b4281a255b46dc1980a084793863e0781b8

bins/stackabled/src/middleware/authentication.rs
nightkr: Actually, I'd like to think a bit harder about how we want to expose errors in general before tackling this, so pulling it out into a separate issue.

nightkr: https://github.com/stackabletech/stackable-cockpit/issues/51

bins/stackabled/src/middleware/authentication.rs
nightkr: a6e8a8e626a65759749fa71d448a1fc5b36622d5

web/src/api/index.ts
nightkr: c605f78ae3faa4228e0a69e2dd09b21a8610246d

Had some issues before, but they turned out to be unrelated to template strings in the end. Only using a single level though, Emacs seems to get confused by nested templates (even if they're technically legal).

web/src/api/index.ts
```diff
@@ -1,7 +1,37 @@
 import createClient from 'openapi-fetch';
 import { components, paths } from './schema';
+import { createLocalStorageSignal } from '../utils/localstorage';
 
 const client = createClient<paths>({ baseUrl: '/api' });
+const [currentSessionToken, setCurrentSessionToken] =
+  createLocalStorageSignal('sessionToken');
+export const isLoggedIn = () => currentSessionToken() !== undefined;
```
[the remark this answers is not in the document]
nightkr: 77c71c9b0a8dc7ba15a599137ab327d9aac176f8

web/src/api/index.ts
nightkr: 436880001989c8518af7ffa78c38b61eaf2a7a9d

web/src/api/index.ts
nightkr: c2600fcf74521bf1eeb9324b03ab401382345397

web/src/pages/login.tsx
nightkr: 917dc69eb1c5f37890c2843c727e2ce532786c2a

web/src/utils/localstorage.ts
nightkr: We don't, the effect is only updated when `value()` changes.

Techassi: Okay then I misinterpreted the function doc comment: `Creates a reactive computation that runs after the render phase`.

nightkr: It runs after the render phase, if it has a reason to :P

web/src/api/index.ts
nightkr: 3c8d5d3bf61eba8e97e57cb2b83887efaf8f03ad

web/src/api/index.ts
Techassi: Mhm bummer. Let's keep it like this and eventually update it.

web/.eslintrc
```diff
@@ -30,6 +30,13 @@
         }
       }
     ],
+    // foo(undefined) is not always equivalent to foo()
+    "unicorn/no-useless-undefined": [
+      "error",
+      {
+        "checkArguments": false
+      }
+    ],
```
Techassi, line 39: I think we don't need this anymore as we removed the undefined stuff by using `Option` right?

web/src/api/session.ts
```diff
@@ -0,0 +1,55 @@
+import openapiCreateClient from 'openapi-fetch';
+import { paths } from './schema';
+import { createLocalStorageSignal } from '../utils/localstorage';
+import { None, someIfDefined } from '../types';
+import { createMemo } from 'solid-js';
+
+const createClient = (options: RequestInit) =>
```
Techassi, line 7: I would make the `option` parameter optional:

```ts
const createClient = (options?: RequestInit = {}) => {}; // or
const createClient = (options: Option<RequestInit>) => {};
```

In the following code we then can use:

```ts
const client = createClient(); // or
const client = createClient(None);
```

web/src/api/session.ts
```diff
@@ -0,0 +1,55 @@
...
+      );
+  }
+}
+
+export async function logIn(
+  username: string,
+  password: string,
+): Promise<string | undefined> {
+  // Always use unauthenticated client for login requests
+  const response = await createClient({}).post('/login', {
```
Techassi, line 43: For example here, see above comment.

Techassi (commented):
LGTM, only minor stuff.

web/src/api/session.ts
nightkr: b219db2896ebce6806b93a06b58b2d14630dfe5a

web/.eslintrc
nightkr: b219db2896ebce6806b93a06b58b2d14630dfe5a

PR #49, opened 2023-07-03
Connect stacklet UI to API

### LINKED ISSUE #14: Show list of stacklets (ZooKeepers, HDFS clusters, etc)

nightkr:
# Description

Fixes #14 

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


web/src/main.css
```diff
@@ -0,0 +1,6 @@
+.inline-list-item {
+  display: inline;
+}
+.inline-list-item:not(:last-child)::after {
+  content: ', ';
+}
```
Techassi, line 6: This can be defined as a shortcut:

```ts
shortcuts: {
  'inline-list-item': 'inline not-last:after:content-[",_"]'
}
```

web/src/main.tsx
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import '@unocss/reset/sanitize/sanitize.css';
 import 'virtual:uno.css';
+import './main.css';
```
Techassi, line 4: With the shortcut, this is not needed anymore.

src/platform/stacklet/mod.rs
```diff
@@ -41,37 +48,20 @@ pub enum StackletError {
     JsonError { source: serde_json::Error },
 }
 
-pub type StackletList = IndexMap<String, Vec<Product>>;
+pub type StackletList = Vec<Stacklet>;
```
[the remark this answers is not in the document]
nightkr: Good catch. I planned to get rid of that but forgot about it.

web/src/main.css
nightkr: I suppose, but I'm also reluctant to add *more* Tailwindisms at this point (#54).

src/platform/stacklet/opensearch.rs
```diff
@@ -25,12 +25,13 @@ pub(super) async fn list_products(
             None => vec![],
         };
 
-        products.push(Product {
+        stacklets.push(Stacklet {
             name: service.name_any(),
             namespace: service.namespace(),
+            product: "opensearch".to_string(),
```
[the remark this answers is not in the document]
nightkr: I thought I copied it across from `mod.rs` but it looks like you're right. Will fix.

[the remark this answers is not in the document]
nightkr: 03e1309de9d3545ee9a639ab7a566fd0481bc09d

src/platform/stacklet/mod.rs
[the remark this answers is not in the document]
nightkr: 583369a51d18aae9af1e957c4c304f47de4fbec4

web/src/main.css
nightkr: Converted it to a CSS module: b04dab6fab7bc49ecab449d3d0e83059fab9d744

Techassi: Okay what are the pain points when using Uno? I guess we can use individual files for nice stuff, but Uno is pretty good at covering 95% of the styles we need.

nightkr: It's "solving" a non-problem. It's adding an extra step of indirection to all docs lookups (instead of `mdn -> code` (or just writing the code directly) everything turns into `mdn -> docs.uno -> code -> docs.uno to understand wtf is going on`) for no benefit. `class` quickly turns into an unmanageable mess, especially since you can't really do comments (without factoring it out of the immediate `class`). The Firefox debugger isn't aware of it, so I still end up writing it as real CSS first *anyway* before then converting it into Tailwind gibberish, so the idea of it even saving any keystrokes in the end is absurd. And it pushes you away from CSS features like grids that can help make your stylesheets significantly more understandable, since they can't really map up to static class sets.

And all of this, just so that you can write `ml-1` instead of `margin-left: 0.25rem;`.

I'm not against abstraction. I'm not against the idea of CSS frameworks. But they have to provide actual value to make up for the costs involved when using them. Tailwind provides none of that. It's just [ProGuard](https://www.guardsquare.com/proguard) for people who thought that that was too automated.

Techassi: > It's adding an extra step of indirection to all docs lookups (instead of mdn -> code (or just writing the code directly) everything turns into mdn -> docs.uno -> code -> docs.uno to understand wtf is going on) for no benefit.

This is not an issue for me personally. I know what I need/want to use 95% of the time. Using these classes *has* a learning curve, but it isn't super steep in my opinion.

> `class` quickly turns into an unmanageable mess, especially since you can't really do comments (without factoring it out of the immediate `class`). 

Agreed, it can messy pretty quick. This is were shortcuts and separation into components helps tho. Plain CSS files can also get super messy (and long).

> And it pushes you away from CSS features like grids that can help make your stylesheets significantly more understandable

How? You can still use all the new fancy layout stuff, like grid, flex, etc...

> I'm not against abstraction. I'm not against the idea of CSS frameworks. But they have to provide actual value to make up for the costs involved when using them. Tailwind provides none of that.

Agreed, and I think Uno *provides* advantages and makes your life easier when writing CSS. I find it quite easy to use them and writing CSS is more costly for me personally (which I did many years, ugh).

> It's just [ProGuard](https://www.guardsquare.com/proguard) for people who thought that that was too automated.

Yes ProGuard is garbage :)

nightkr: > This is not an issue for me personally. I know what I need/want to use 95% of the time. Using these classes has a learning curve, but it isn't super steep in my opinion.

And, again, what do you gain in exchange for this extra obfuscation?

> Agreed, it can get messy pretty quick. This is where shortcuts and separation into components helps tho. Plain CSS files can also get super messy (and long).

And that's an improvement over plain old CSS.. how?

> How? You can still use all the new fancy layout stuff, like grid, flex, etc...

https://unocss.dev/interactive/?s=grid-template

Techassi: > And, again, what do you gain in exchange for this extra obfuscation?

Maintainability, unified CSS code style, unified access for sizing, colors, spacing, no unused CSS rules, etc...

> And that's an improvement over plain old CSS.. how?

No nesting, obnoxious variable declaration and usage, possible multiple implementations of the same rules, hundreds or even thousands of lines of code across many files.

> https://unocss.dev/interactive/?s=grid-template

You can still use `grid-cols-<num>` and `grid-rows-<num>`. Tbh, I never needed nor used grid areas.

nightkr: > no unused CSS rules

[I've got a bridge to sell you](https://unpkg.com/tailwindcss@2.2.19/dist/tailwind.css)

Techassi: All these rules won't even be generated. The current generated CSS is only 110 rules with a size of 7.5 kB (not minified).

nightkr: The rules are *effectively* there and need to be taken into account. (Of course, this is only a tiny fraction of the infinite number of rules that actually exist once you take Uno's dynamic rules into account.)

Techassi: I really don't get what you are trying to tell me here. All generated CSS is based on the classes you _actually_ use, not those which _could_ be there if you used all possible class combinations. Why should we care about **ALL** possible rules, if we don't use them and thus will not be generated?

If I manually maintain CSS rules, I can still just re-declare the same set of rules using different selectors. I could write the same block of rules 10 times and each time I use a different selector both in the CSS and the HTML (TSX) file. This bloat basically is only discovered if you go over the CSS with a specialized tool or manually look for these kind of issues and fix them yourself.

nightkr: > Why should we care about ALL possible rules, if we don't use them and thus will not be generated?

We still have to act as if the names exist and are reserved by the framework.

> If I manually maintain CSS rules, I can still just re-declare the same set of rules using different selectors. I could write the same block of rules 10 times and each time I use a different selector both in the CSS and the HTML (TSX) file. This bloat basically is only discovered if you go over the CSS with a specialized tool or manually look for these kind of issues and fix them yourself.

And you can declare identical TSX components ten times over too, I don't see how that's an argument one way or another.

Besides, I'd argue that having identical rules isn't a problem as long as they have separate semantic intents.

web/src/main.tsx
Techassi: Resolved, we use SCSS in the future.

web/src/main.css
Techassi: Resolved, we use SCSS in the future.

Techassi (approved):

PR #73, opened 2023-07-31
Basic translation support using Fluent

### LINKED ISSUE #64: Integrate i18n into frontend

nightkr:
# Description

Fixes #64 

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/web/src/localization/index.tsx
+++ b/web/src/localization/index.tsx
@@ -30,7 +30,7 @@ const translationBundles = Object.fromEntries(
   }),
 );
 
-const fallbackLocale = 'en';
+const FALLBACK_LOCALE = 'en';
 interface LanguageContext {
   activeLocales(): string[];
   setUserLocale(language: Option<string>): void;
@@ -57,10 +57,10 @@ export const LanguageProvider = (props: { children: JSX.Element }) => {
           navigator.languages,
           Object.keys(translationBundles),
           {
-            defaultLocale: fallbackLocale,
+            defaultLocale: FALLBACK_LOCALE,
           },
         ),
-      (ul) => [ul, fallbackLocale],
+      (ul) => [ul, FALLBACK_LOCALE],
     ),
   );
   return (


web/src/localization/locale/en.ftl
Techassi, line 1: We can also add the German translations:

```
product = Produkt
namespace = Namensraum
name = Name
status = Status

stacklets-list = Stacklets
stacklets-add = Stacklet hinzufügen

log-out = Abmelden

refresh = Aktualisieren
```

nightkr: 9a43e2319ec7f5e241f81b4078cc948e4433757b

web/src/components/datatable.tsx
```diff
@@ -57,7 +58,9 @@ export function DataTable<T>(props: DataTableProps<T>): JSX.Element {
         <div class='flex-grow' />
         {props.extraButtons}
         <Show when={props.refresh}>
-          <Button onClick={() => props.refresh?.()}>Refresh</Button>
+          <Button onClick={() => props.refresh?.()}>
+            {translate('refresh')}
+          </Button>
```
Techassi, line 63: Might be worth to handle translation inside the button it self. This way we also force us to add translations for new button actions / texts.

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const fallbackLanguage = 'en';
+const LanguageContext = createContext<
+  [
+    Accessor<string[]>,
+    {
+      setUserLanguage: (language: Option<string>) => void;
+      availableLanguages: () => { [language: string]: FluentBundle };
+    },
+  ]
```
Techassi, line 36: This type can be extracted instead of defining this inline.

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+>();
+export const requireLanguageContext = () => {
+  const langContext = useContext(LanguageContext);
+  if (langContext == undefined) {
```
Techassi, line 40: This should be sufficient:

 ```suggestion
  if (!langContext) {
```

File: web/src/localization/index.tsx [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+      ]}
+    >
+      {props.children}
+    </LanguageContext.Provider>
+  );
+};
+
+export const translate = (
+  name: string,
+  variables?: { [variable: string]: string | number },
```
Techassi, line 74: Let's just mirror the internal Fluent type:

```suggestion
  variables?: Record<string, FluentVariable>,
```

Needs import:

```ts
import { FluentVariable } from '@fluent/bundle';
```

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+    console.error(
+      `No translation found for ${name} (language chain: ${languageChain.toString()})`,
+    );
+    return `#${name}#`;
+  }
+  const pattern = bundle.getMessage(name)?.value;
+  if (pattern == undefined) {
+    throw new Error('Translation pattern for ${name} has no value');
+  }
+  return bundle.formatPattern(pattern, variables);
```
Techassi, line 93: Let's space this out

web/src/localization/index.tsx
Techassi, line 1: We might opt for a slightly shorter (but still recognized) folder name: `i18n`.

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      throw new Error(
+        `Failed to load translation bundle ${fileName} as ${languageName}`,
+        { cause: errors[0] },
+      );
+    }
+    return [languageName, bundle];
+  }),
+);
+
+const fallbackLanguage = 'en';
```
Techassi, line 28: We might want to use uppercase here.

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+    name,
+  );
+  if (bundle == undefined) {
+    console.error(
+      `No translation found for ${name} (language chain: ${languageChain.toString()})`,
+    );
+    return `#${name}#`;
+  }
+  const pattern = bundle.getMessage(name)?.value;
+  if (pattern == undefined) {
```
Techassi, line 90: Let's use:

```suggestion
  if (!pattern) {
```

Techassi (changes requested):
Only minor stuff.

web/src/components/datatable.tsx
nightkr: That creates a whole bunch of ambiguities, 1) not everything will be in a special purpose container, 2) we'd need to keep track of which containers force i18n and which don't, 3) i18nizing containers become "terminal" and can't be composed further.

I also kind of like that this makes it easy to grep for `translate(` if you want to find all translation strings (probably something we'll want to look tooling for at some point to track).

web/src/localization/locale/en.ftl
```diff
@@ -0,0 +1,17 @@
+language-name = English
+
+refresh = Refresh
+
+stacklet--list = Stacklets
```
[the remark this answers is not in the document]
Techassi: OKay that's what I thought. Thank you for the explanation!

web/src/localization/locale/de.ftl
```diff
@@ -0,0 +1,14 @@
...
+stacklet--list = Stacklets
+stacklet--add = Stacklet hinzufügen
+
+stacklet--product = Produkt
+stacklet--namespace = Namensraum
+stacklet--name = Name
+stacklet--status = Status
+
+
+login--log-out = Abmelden
```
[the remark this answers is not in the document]
nightkr: 8c90142f0156887831c90fc15147a8344ed356cb

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
+import { FluentBundle, FluentResource } from '@fluent/bundle';
+import { mapBundleSync } from '@fluent/sequence';
+import { negotiateLanguages } from '@fluent/langneg';
+import { Accessor, JSX, createContext, createMemo, useContext } from 'solid-js';
+import { createLocalStorageSignal } from '../utils/localstorage';
+import { Option } from '../types';
+
+const ftls = import.meta.glob('./locale/*.ftl', { eager: true, as: 'raw' });
```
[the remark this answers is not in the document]
nightkr: c4271ca60d171477b8f16eccd278cddd42ecd5ad

web/src/localization/index.tsx
nightkr: c4271ca60d171477b8f16eccd278cddd42ecd5ad

web/src/localization/index.tsx
nightkr: c4271ca60d171477b8f16eccd278cddd42ecd5ad

web/src/localization/index.tsx
nightkr: 12db52f817ab43d2ed1e71ae0d59d2789a5a1569

web/src/localization/index.tsx
nightkr: Hm?

web/src/localization/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const errors = bundle.addResource(resource);
+    if (errors.length > 0) {
+      throw new Error(
+        `Failed to load translation bundle ${fileName} as ${languageName}`,
+        { cause: errors[0] },
+      );
+    }
+    return [languageName, bundle];
+  }),
+);
```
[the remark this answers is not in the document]
nightkr: 12db52f817ab43d2ed1e71ae0d59d2789a5a1569

web/src/localization/index.tsx
Techassi: That's true. I think the current setup of ESLint won't force you to explicitly compare against `undefined`. If we want to explicitly use `undefined`, I would like us to use `===` instead of `==`.

web/src/localization/index.tsx
Techassi: I think it makes sense to use `FALLBACK_LANGUAGE` instead of `fallbackLanguage` for the variable name.

web/src/localization/index.tsx
Techassi: Marking this as resolved. See discussion above.

web/src/localization/index.tsx
nightkr: Sure, a432e0c47876029cafc36a0da3b421c8e4feffb9

web/src/localization/index.tsx
nightkr: 21837f6b748f837f162bf7c3ffcdaa482e5ebfcd

Techassi: Great choice of a commit title :)

web/src/components/datatable.tsx
Techassi: Your points are solid. Let's use the current approach.

web/src/localization/index.tsx
Techassi: Do we want to change that? Or keep the current folder name? I don't have a super strong opinion on that.

Techassi (approved):
Only one minor discussion left. Approved.

web/src/localization/index.tsx
nightkr: I don't really care either way. I agree that the shorter folder name should still be recognizable but have been trying to avoid names that would upset our poor dear eslint bot (even if AFAIK it doesn't actually check folder names, it's probably nice to be consistent).

Personally I'd prefer to be a bit looser on abbreviations in general (as long as they're common and unambiguous in the context).

nightkr: Merging for now, I don't think it's worth overthinking this and we can always revisit later.

Techassi: ESLint only checks file names - you can also add your own allowed names if ESLint complains.

So I think renaming to `i18n` should be fine :)

nightkr: It's more about the principle of the thing. :P

PR #76, opened 2023-08-01
feat: Add cache auto-purge and `--old` arg to cache clean command

Techassi:
This PR adds a cache auto-purge feature, which cleans up cached files which are outdated. This is run even if the operation doesn't need cached (or remote) files. This helps to keep the cache folder clean.

The PR also adds `--old / --outdated` to the `stackablectl cache clean` command to only clear cached files which are outdated. Before this change, all cached files (even files not considered outdated) were deleted.

### DIFF
--- a/rust/stackable-cockpit/src/xfer/cache.rs
+++ b/rust/stackable-cockpit/src/xfer/cache.rs
@@ -31,9 +31,6 @@ pub enum Error {
 
     #[snafu(display("tried to write file with disabled cache"))]
     WriteDisabled,
-
-    #[snafu(display("failed to convert OsString into string"))]
-    OsStringConvertError,
 }
 
 #[derive(Debug)]
@@ -113,7 +110,7 @@ impl Cache {
                     let metadata = entry.metadata().await.context(CacheIoSnafu)?;
 
                     // Skip protected files
-                    if is_protected_file(entry.file_name())? {
+                    if is_protected_file(entry.file_name()) {
                         continue;
                     }
 
@@ -137,8 +134,8 @@ impl Cache {
                     let metadata = entry.metadata().await.context(CacheIoSnafu)?;
 
                     let should_delete_file = match delete_filter {
-                        // Skip the last-auto-purge file
-                        _ if is_protected_file(entry.file_name())? => false,
+                        // Skip protected files
+                        _ if is_protected_file(entry.file_name()) => false,
 
                         // Without --old / --outdated
                         DeleteFilter::All => true,
@@ -168,6 +165,16 @@ impl Cache {
         match &self.backend {
             CacheBackend::Disk { base_path } => {
                 let cache_auto_purge_filepath = base_path.join(CACHE_LAST_AUTO_PURGE_FILEPATH);
+
+                // Only create file if not already present
+                if !fs::try_exists(&cache_auto_purge_filepath)
+                    .await
+                    .context(CacheIoSnafu)?
+                {
+                    write_cache_auto_purge_file(&cache_auto_purge_filepath).await?;
+                }
+
+                // Read and covert timestamp
                 let timestamp = fs::read_to_string(&cache_auto_purge_filepath)
                     .await
                     .context(CacheIoSnafu)?;
@@ -175,11 +182,13 @@ impl Cache {
                 let timestamp: u64 = timestamp.parse().context(ParseIntSnafu)?;
                 let timestamp = UNIX_EPOCH + Duration::from_secs(timestamp);
 
+                // If the auto purge interval elapsed, run purge and write
+                // back the new timestamp
                 if timestamp.elapsed().context(SystemTimeSnafu)? >= self.auto_purge_interval {
                     debug!("Auto-purging outdated cache files");
 
                     self.purge(DeleteFilter::OnlyExpired).await?;
-                    write_cache_auto_purge_file(cache_auto_purge_filepath).await?;
+                    write_cache_auto_purge_file(&cache_auto_purge_filepath).await?;
                 }
 
                 Ok(())
@@ -262,15 +271,6 @@ impl CacheSettings {
         match &self.backend {
             CacheBackend::Disk { base_path } => {
                 fs::create_dir_all(base_path).await.context(CacheIoSnafu)?;
-                let cache_auto_purge_filepath = base_path.join(CACHE_LAST_AUTO_PURGE_FILEPATH);
-
-                // Only create file if not already present
-                if !fs::try_exists(&cache_auto_purge_filepath)
-                    .await
-                    .context(CacheIoSnafu)?
-                {
-                    write_cache_auto_purge_file(cache_auto_purge_filepath).await?;
-                }
 
                 Ok(Cache::new(
                     self.backend,
@@ -308,7 +308,7 @@ impl From<bool> for DeleteFilter {
     }
 }
 
-async fn write_cache_auto_purge_file(path: PathBuf) -> Result<()> {
+async fn write_cache_auto_purge_file(path: &Path) -> Result<()> {
     fs::write(
         path,
         SystemTime::now()
@@ -322,11 +322,8 @@ async fn write_cache_auto_purge_file(path: PathBuf) -> Result<()> {
     .context(CacheIoSnafu)
 }
 
-fn is_protected_file(filename: OsString) -> Result<bool> {
-    Ok(CACHE_PROTECTED_FILES.contains(
-        &filename
-            .into_string()
-            .map_err(|_| Error::OsStringConvertError)?
-            .as_str(),
-    ))
+fn is_protected_file(filename: OsString) -> bool {
+    // Non-UTF-8 filenames can't possibly be on the protected list
+    let Some(filename) = filename.to_str() else { return false; };
+    CACHE_PROTECTED_FILES.contains(&filename)
 }
--- a/rust/stackable-cockpit/src/xfer/cache.rs
+++ b/rust/stackable-cockpit/src/xfer/cache.rs
@@ -166,25 +166,21 @@ impl Cache {
             CacheBackend::Disk { base_path } => {
                 let cache_auto_purge_filepath = base_path.join(CACHE_LAST_AUTO_PURGE_FILEPATH);
 
-                // Only create file if not already present
-                if !fs::try_exists(&cache_auto_purge_filepath)
-                    .await
-                    .context(CacheIoSnafu)?
-                {
-                    write_cache_auto_purge_file(&cache_auto_purge_filepath).await?;
-                }
-
                 // Read and covert timestamp
-                let timestamp = fs::read_to_string(&cache_auto_purge_filepath)
-                    .await
-                    .context(CacheIoSnafu)?;
-
-                let timestamp: u64 = timestamp.parse().context(ParseIntSnafu)?;
-                let timestamp = UNIX_EPOCH + Duration::from_secs(timestamp);
+                let last_purged_at = match fs::read_to_string(&cache_auto_purge_filepath).await {
+                    Ok(timestamp) => Some(
+                        UNIX_EPOCH + Duration::from_secs(timestamp.parse().context(ParseIntSnafu)?),
+                    ),
+                    Err(err) if err.kind() == std::io::ErrorKind::NotFound => None,
+                    Err(err) => return Err(err).context(CacheIoSnafu),
+                };
 
                 // If the auto purge interval elapsed, run purge and write
                 // back the new timestamp
-                if timestamp.elapsed().context(SystemTimeSnafu)? >= self.auto_purge_interval {
+                if last_purged_at
+                    .and_then(|ts| ts.elapsed().ok())
+                    .map_or(true, |elapsed| elapsed >= self.auto_purge_interval)
+                {
                     debug!("Auto-purging outdated cache files");
 
                     self.purge(DeleteFilter::OnlyExpired).await?;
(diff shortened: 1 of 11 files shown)


rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -20,12 +25,19 @@ pub enum Error {
     #[snafu(display("system time error"))]
     SystemTimeError { source: SystemTimeError },
 
+    #[snafu(display("failed to parse string as integer"))]
+    ParseIntError { source: ParseIntError },
```
nightkr, line 29: There error messages are nigh-useless, _what_ are you trying to parse? Where should the user look to diagnose the issue?

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -97,15 +109,14 @@ impl Cache {
...
-                            .metadata()
-                            .await
-                            .context(CacheIoSnafu)?
-                            .modified()
-                            .context(CacheIoSnafu)?,
-                    ))
+                    let metadata = entry.metadata().await.context(CacheIoSnafu)?;
+
+                    // Skip the last-auto-purge file
+                    if metadata.is_file() && is_protected_file(entry.file_name())? {
```
nightkr, line 115: Is there a reason to limit this to regular files?

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -97,15 +109,14 @@ impl Cache {
...
+                        continue;
+                    }
+
+                    files.push((entry.path(), metadata.modified().context(CacheIoSnafu)?))
                 }
 
                 files.sort();
```
nightkr, line 122: Unrelated: does anything rely on this sorting? Why do we do it here rather than where that assumption exists?

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -117,18 +128,70 @@ impl Cache {
 
     /// Removes all cached files by deleting the base cache folder and then
     /// recreating it.
-    pub async fn purge(&self) -> Result<()> {
+    pub async fn purge(&self, only_old_files: bool) -> Result<()> {
```
nightkr, line 131: Consider making this an `enum` instead:

```rust
enum DeleteFilter {
  All,
  OnlyExpired,
}
```

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -117,18 +128,70 @@ impl Cache {
...
+                            .elapsed()
+                            .context(SystemTimeSnafu)?
+                            > self.max_age
+                    {
+                        fs::remove_file(entry.path()).await.context(CacheIoSnafu)?;
+                        continue;
+                    }
+
+                    // Without --old / --outdated
+                    fs::remove_file(entry.path()).await.context(CacheIoSnafu)?;
```
nightkr, line 158:
```suggestion
                    let should_delete_file = match filter {
                      // Skip the last-auto-purge file
                      _ if is_protected_file(entry.file_name())? => false,
                      
                      DeleteFilter::All => true,
                      // with --old/--outdated
                      DeleteFilter::OnlyExpired => metadata
                            .modified()
                            .context(CacheIoSnafu)?
                            .elapsed()
                            .context(SystemTimeSnafu)?
                            > self.max_age,
                    };
                    if should_delete_file {
                        fs::remove_file(entry.path()).await.context(CacheIoSnafu)?;
                    }
```

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -117,18 +128,70 @@ impl Cache {
...
+                let cache_auto_purge_filepath = base_path.join(CACHE_LAST_AUTO_PURGE_FILEPATH);
+                let timestamp = fs::read_to_string(&cache_auto_purge_filepath)
+                    .await
+                    .context(CacheIoSnafu)?;
+
+                let timestamp: u64 = timestamp.parse().context(ParseIntSnafu)?;
+                let timestamp = UNIX_EPOCH + Duration::from_secs(timestamp);
+
+                if timestamp.elapsed().context(SystemTimeSnafu)? >= self.auto_purge_interval {
+                    self.purge(true).await?;
```
nightkr, line 179:
```suggestion
                    tracing::debug!("auto-purging");
                    self.purge(true).await?;
```

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -195,9 +260,27 @@ impl CacheSettings {
...
-                Ok(Cache::new(self.backend, self.max_age))
+                let cache_auto_purge_filepath = base_path.join(CACHE_LAST_AUTO_PURGE_FILEPATH);
+
+                // Only create file if not already present
+                if !fs::try_exists(&cache_auto_purge_filepath)
+                    .await
+                    .context(CacheIoSnafu)?
+                {
+                    write_cache_auto_purge_file(cache_auto_purge_filepath).await?;
+                }
```
nightkr, line 271: Isn't it enough for `auto_purge()` to maintain this?

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -207,3 +290,26 @@ pub enum CacheBackend {
     Disk { base_path: PathBuf },
     Disabled,
 }
+
+async fn write_cache_auto_purge_file(path: PathBuf) -> Result<()> {
```
nightkr, line 294:
```suggestion
async fn write_cache_auto_purge_file(path: &Path) -> Result<()> {
```

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -207,3 +290,26 @@ pub enum CacheBackend {
...
+}
+
+fn is_protected_file(filename: OsString) -> Result<bool> {
+    Ok(CACHE_PROTECTED_FILES.contains(
+        &filename
+            .into_string()
+            .map_err(|_| Error::OsStringConvertError)?
+            .as_str(),
+    ))
+}
```
nightkr, line 315:
```suggestion
fn is_protected_file(filename: &OsStr) -> bool {
    // non-UTF-8 filenames can't possibly be on the protected list
    let Some(filename) = filename.to_str() else { return false; };
    CACHE_PROTECTED_FILES.contains(filename)
}
```

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Agreed. Tbh I was just a little lazy to come up with a better message :)

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Not really. We can remove this check to allow protected dirs in the future (if ever needed).

Techassi: Removed in b51571d

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Removed in b51571d

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Added in b51571d

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Changed in b51571d

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Added in b51571d

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Changed in 2eaaa8b

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Changed in 2eaaa8b

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -291,6 +293,21 @@ pub enum CacheBackend {
...
+
+impl From<bool> for DeleteFilter {
+    fn from(value: bool) -> Self {
+        if value {
+            Self::OnlyExpired
+        } else {
+            Self::All
+        }
+    }
+}
```
nightkr, line 309: Please be judicious about when you implement `From<T>`. For `bool` especially, you'll often need a lot of additional context to understand what the field actually means. Instead, either introduce a separate named associated function, or just inline the conversion to the one place where it's actually used.

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -128,34 +128,34 @@ impl Cache {
...
-                    if only_old_files
-                        && metadata
-                            .modified()
-                            .context(CacheIoSnafu)?
-                            .elapsed()
-                            .context(SystemTimeSnafu)?
-                            > self.max_age
-                    {
+                    let should_delete_file = match delete_filter {
+                        // Skip the last-auto-purge file
```
nightkr:
```suggestion
                        // Skip protected files
```
Might as well align with L115.

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: `auto_purge()` now maintains this file since 2eaaa8b

rust/stackable-cockpit/src/xfer/cache.rs
nightkr: Already done, heh.

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Removed, not needed anymore. 2eaaa8b

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Yeah. I had similar thoughts. Will add an `from_bool` method to `DeleteFilter`.

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Done in 2eaaa8b

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Added in f7b0df5

rust/stackable-cockpit/src/xfer/cache.rs
```diff
@@ -168,18 +165,30 @@ impl Cache {
...
 
+                // If the auto purge interval elapsed, run purge and write
+                // back the new timestamp
                 if timestamp.elapsed().context(SystemTimeSnafu)? >= self.auto_purge_interval {
                     debug!("Auto-purging outdated cache files");
 
                     self.purge(DeleteFilter::OnlyExpired).await?;
-                    write_cache_auto_purge_file(cache_auto_purge_filepath).await?;
+                    write_cache_auto_purge_file(&cache_auto_purge_filepath).await?;
                 }
```
nightkr: This can be simplified a bit if we say that it's ok to purge when there is no timestamp or there have been time travel shenanigans:

```rust
// Read and covert timestamp
let last_purged_at = match fs::read_to_string(&cache_auto_purge_filepath)
    .await {
        Ok(timestamp) => Some(UNIX_EPOCH + Duration::from_secs(timestamp.parse().context(ParseIntSnafu)?)),
        Err(err) if err.kind() == std::io::ErrorKind::NotFound => None,
        Err(err) => return Err(err).context(CacheIoSnafu),
    }
    .context(CacheIoSnafu)?;

// If the auto purge interval elapsed, run purge and write
// back the new timestamp
if last_purged_at.and_then(|ts| ts.elapsed().ok()).map_or(true, |elapsed| elapsed >= self.auto_purge_interval) {
    debug!("Auto-purging outdated cache files");

    self.purge(DeleteFilter::OnlyExpired).await?;
    write_cache_auto_purge_file(&cache_auto_purge_filepath).await?;
}
```

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: What name do you suggest?

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Also thought about this approach. Will add it.

rust/stackable-cockpit/src/xfer/cache.rs
nightkr: I'd just inline it, since it's only really useful in the context of the CLI. Alternatively, we could implement `clap::Parser` on `DeleteFilter`, but honestly that seems a bit overkill.

Techassi: > Alternatively, we could implement clap::Parser on DeleteFilter, but honestly that seems a bit overkill.

My thoughts as well. Let's inline it I guess...

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Added in a2a4dc1

rust/stackable-cockpit/src/xfer/cache.rs
Techassi: Inlined in a2a4dc1

nightkr (approved):

PR #79, opened 2023-08-03
feat(ctl): Namespace separation and support skipping release installation

Techassi:
Fixes https://github.com/stackabletech/stackablectl/issues/259
Needs https://github.com/stackabletech/stackablectl/pull/274


rust/stackable-cockpit/src/constants.rs
```diff
@@ -3,8 +3,9 @@ use std::time::Duration;
 pub const REDACTED_PASSWORD: &str = "<redacted>";
 pub const PASSWORD_LEN: usize = 32;
 
-pub const DEFAULT_STACKABLE_NAMESPACE: &str = "stackable";
-pub const DEFAULT_NAMESPACE: &str = "default";
+pub const DEFAULT_OPERATOR_NAMESPACE: &str = "stackable-operators";
+// TODO (Techassi): Change this to "stackable-demos" once we switch to this version
```
sbernauer, line 7:
```suggestion
// TODO (Techassi): Change this to "stackable-demos" once we switch to this version.
// Currently lots of demos can only run in the default namespace, so we have to keep
// "default" here, until we switch the demos. We cant switch them right now, as the old
// stackablectl would break
```

rust/stackable-cockpit/src/constants.rs
```diff
@@ -3,8 +3,9 @@ use std::time::Duration;
...
+pub const DEFAULT_PRODUCT_NAMESPACE: &str = "default";
```
sbernauer, line 8: Question is what we want to call the namespace. I would be in favor of `stackable`, as there could be stacks or demos in there

rust/stackablectl/src/args/namespace.rs
```diff
@@ -0,0 +1,14 @@
...
+use stackable_cockpit::constants::{DEFAULT_OPERATOR_NAMESPACE, DEFAULT_PRODUCT_NAMESPACE};
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Namespace options")]
+pub struct CommonNamespaceArgs {
+    /// Namespace in the cluster used to deploy the operators
+    #[arg(short = 'n', long, default_value = DEFAULT_OPERATOR_NAMESPACE, global = true)]
+    pub operator_namespace: String,
+
+    /// Namespace in the cluster used to deploy the products
```
sbernauer, line 11: Applies to all occurrences: I think this term is more general and makes more sense on a list command
```suggestion
    /// Namespace where the products (e.g. stacks or demos) are deployed
```

rust/stackablectl/src/args/namespace.rs
```diff
@@ -0,0 +1,14 @@
+use clap::Args;
+use stackable_cockpit::constants::{DEFAULT_OPERATOR_NAMESPACE, DEFAULT_PRODUCT_NAMESPACE};
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Namespace options")]
+pub struct CommonNamespaceArgs {
+    /// Namespace in the cluster used to deploy the operators
+    #[arg(short = 'n', long, default_value = DEFAULT_OPERATOR_NAMESPACE, global = true)]
```
sbernauer, line 8: Please use the `short = 'n'` for the product namespaces

rust/stackablectl/src/args/repo.rs
```diff
@@ -0,0 +1,19 @@
+use clap::{Args, ValueHint};
+
+use crate::constants::{HELM_REPO_URL_DEV, HELM_REPO_URL_STABLE, HELM_REPO_URL_TEST};
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Helm repository options")]
+pub struct CommonRepoArgs {
+    // Provide a custom Helm stable repository URL
```
sbernauer, line 8:
```suggestion
    /// Provide a custom Helm stable repository URL
```

rust/stackablectl/src/cli/mod.rs
```diff
@@ -213,7 +140,7 @@ pub enum Commands {
 
     /// Interact with deployed stacklets, which are bundles of resources and
     /// containers required to run the product.
-    #[command(aliases(["stl", "sl"]))]
+    #[command(aliases(["stl", "sl", "svc"]))]
```
sbernauer, line 143: Wuhu :D I think I'm also happy to remove it

rust/stackablectl/src/args/namespace.rs
```diff
@@ -0,0 +1,14 @@
+use clap::Args;
+use stackable_cockpit::constants::{DEFAULT_OPERATOR_NAMESPACE, DEFAULT_PRODUCT_NAMESPACE};
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Namespace options")]
+pub struct CommonNamespaceArgs {
+    /// Namespace in the cluster used to deploy the operators
```
sbernauer, line 7: Applies to all occurrences: I think this term is more general and makes more sense on a list command
```suggestion
    /// Namespace where the operators are deployed
```

rust/stackablectl/src/cmds/operator.rs
```diff
@@ -86,6 +89,10 @@ Use \"stackablectl operator list\" to list available versions for all operators
 Use \"stackablectl operator describe <OPERATOR>\" to get available versions for one operator")]
     operators: Vec<OperatorSpec>,
 
+    /// Namespace in the cluster used to deploy the operators
+    #[arg(short = 'n', long, default_value = DEFAULT_OPERATOR_NAMESPACE)]
```
sbernauer, line 93: Here the -n makes sense to point to operators i think :+1:

sbernauer (changes requested):
Not tested it yet, waiting for non-draft PR with that

rust/stackable-cockpit/src/constants.rs
Techassi: Agreed. I will update the comment.

rust/stackablectl/src/args/namespace.rs
Techassi: Done in 7066caa

rust/stackablectl/src/cmds/operator.rs
Techassi: Removed in 7066caa

rust/stackablectl/src/cli/mod.rs
Techassi: I don't have a strong preference. We can keep it in for some time (to help users migrate) and drop it eventually in the future.

nightkr: If it's incompatible anyway then I don't think there's much value in keeping it.

sbernauer (commented):
Haven't had the time to actually use and test it yet :/
The final missing pice from https://github.com/stackabletech/stackablectl/issues/259 would be 
> Demos have the possibility to say "i can only run in stackable-demo" namespace (in stacks.yaml or demo.yaml). If they do so, a warning is printed when they are installed in a different namespace, saying the demo will likely not work. This can be the case as we have to use FQDN service names so that the TLS certs are valid

We could do it in a follow-up PR but I think it should not be that complicated

rust/stackablectl/src/cli/mod.rs
Techassi: Removed in 07c786d

rust/stackablectl/README.md
```diff
@@ -35,34 +35,36 @@ Options:
       --offline
           Do not request any remote files via the network
 
-  -n, --operator-namespace <OPERATOR_NAMESPACE>
```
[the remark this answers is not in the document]
Techassi: This output is auto-generated. Also this option is removed from the top level. All namespace related options are now only available for commands which need those.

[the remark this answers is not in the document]
sbernauer: auto-generated by git pre-commit hook?

rust/stackable-cockpit/src/platform/namespace.rs
```diff
@@ -0,0 +1,19 @@
+use crate::kube::{KubeClient, KubeClientError};
+
+/// Creates a namespace with `name` if needed (not already present in the
+/// cluster).
+pub async fn create_if_needed(name: String) -> Result<(), KubeClientError> {
+    let client = KubeClient::new().await?;
+    let namespaces = client.list_namespaces().await?;
+
+    let exists = namespaces.iter().any(|ns| match &ns.metadata.name {
```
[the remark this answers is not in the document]
Techassi: Great catch. I will update the `create_if_needed` function.

rust/stackablectl/README.md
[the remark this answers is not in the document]
Techassi: Yes via a pre-commit hook. This usage output is always up to date :)

rust/stackable-cockpit/src/platform/namespace.rs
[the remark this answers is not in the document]
Techassi: Updated in ba4d4cd

sbernauer (changes requested):
```
cargo r -p stackablectl -- stack in trino-superset-s3 -c kind
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/stackablectl stack in trino-superset-s3 -c kind`
Error: stack command error

  1: stack error
  2: release install error *
  3: failed with Helm error *
  4: failed to install Helm release *
  5: helm error: create: failed to create: namespaces "stackable-operators" not found

NOTE: Some redundant information has been removed from the lines marked with *. Set SNAFU_RAW_ERROR_MESSAGES=1 to disable this behavior.
```

rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -40,3 +51,76 @@ pub struct DemoSpecV2 {
...
+
+    #[snafu(display("cannot install demo in namespace '{requested}', only {} supported", supported.join(", ")))]
+    UnsupportedNamespace {
+        requested: String,
+        supported: Vec<String>,
+    },
+}
+
+impl DemoSpecV2 {
+    #[allow(clippy::too_many_arguments)]
```
Techassi, line 71: This needs to be addressed. However, I'm not quite sure how. Do we wrap the arguments in a new type or do we create some kind of builder..?

sbernauer:
Change LGTM, did play around and bit and gave feedback via PN

rust/stackable-cockpit/src/platform/demo/spec.rs
Techassi: Let's leave it like this and re-visit it in the future.

sbernauer (approved):
Whup whup

sbernauer (changes requested):
The only thing missing from https://github.com/stackabletech/stackablectl/issues/259 is:

> stackablectl tries to create specified namespaces if not existing
> *  If this fails because of missing permissions an error is thrown as the user is suggested to specify operators and demo namespace manually

sbernauer (approved):
Thanks for working on this!

sbernauer (approved):

PR #85, opened 2023-08-07
feat(ctl): Support `--helm-repo-multiarch`

Techassi:
This PR adds support for `--helm-repo-multiarch` until we finalized multi arch support and how we want to integrate support for it in `stackablectl`.

~This PR is blocked because another PR (#79) restructures the CLI arguments.~ Not blocked anymore, because #79 is merged.


sbernauer:
Just curious what the outcome of this PR should be ;) Currently it only adds a helm repo but never uses it

Techassi:
Yes currently it doesn't get used anywhere, that's why the PR is still a draft. Eventually you will be able to use this repo just like the current ones (stable, test and dev).

Yes, in theory you could "abuse" some other repo, but I feel like this should be a separate argument.

sbernauer:
Alright, thanks!

PR #95, opened 2023-08-18
refactor: Replace `xdg` with `directories`

Techassi:
This is a drop-in replacement. Using `directories` enables us to build `stackablectl` for Windows as `xdg` didn't support Windows.

### DIFF
--- a/rust/stackablectl/src/cli/mod.rs
+++ b/rust/stackablectl/src/cli/mod.rs
@@ -195,8 +195,6 @@ pub enum CacheSettingsError {
     UserDir,
 }
 
-pub struct InheritStackDemoArgs {}
-
 /// Returns a list of paths or urls based on the default (remote) file and
 /// files provided via the env variable.
 fn get_files(default_file: &str, env_key: &str) -> Result<Vec<PathOrUrl>, PathOrUrlParseError> {
--- a/rust/stackablectl/src/cli/mod.rs
+++ b/rust/stackablectl/src/cli/mod.rs
@@ -1,7 +1,8 @@
 use std::env;
 
 use clap::{Parser, Subcommand, ValueEnum};
-use snafu::{ResultExt, Snafu};
+use directories::ProjectDirs;
+use snafu::Snafu;
 use tracing::{debug, instrument, Level};
 
 use stackable_cockpit::{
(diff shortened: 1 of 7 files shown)


rust/stackablectl/src/constants.rs
```diff
@@ -15,4 +15,6 @@ pub const HELM_REPO_URL_STABLE: &str = "https://repo.stackable.tech/repository/h
 pub const HELM_REPO_URL_TEST: &str = "https://repo.stackable.tech/repository/helm-test/";
 pub const HELM_REPO_URL_DEV: &str = "https://repo.stackable.tech/repository/helm-dev/";
 
-pub const CACHE_HOME_PATH: &str = "stackablectl";
+pub const XDG_APPLICATION_NAME: &str = "stackablectl";
+pub const XDG_ORGANIZATION_NAME: &str = "Stackable";
+pub const XDG_QUALIFIER: &str = "tech";
```
sbernauer, line 20: Should we still prefix this stuff `XDG_*`, when xdg is only used in linux? Or maybe call it e.g. `CACHE_APPLICATION_NAME` or similar

rust/stackablectl/src/cli/mod.rs
```diff
@@ -185,7 +189,7 @@ pub enum OutputType {
 #[snafu(module)]
 pub enum CacheSettingsError {
     #[snafu(display("unable to resolve XDG directories"))]
-    Xdg { source: xdg::BaseDirectoriesError },
+    Xdg,
```
sbernauer, line 192: Should we still call thisxdg, when xdg is only used in linux?

sbernauer (changes requested):

rust/stackablectl/src/constants.rs
Techassi: Renamed in 1fe6c05

rust/stackablectl/src/cli/mod.rs
Techassi: Renamed in 1fe6c05

sbernauer (approved):
Thx!

sbernauer:
Pipeline fails with
```
 error[rejected]: failed to satisfy license requirements
  ┌─ option-ext 0.2.0 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
4 │ license = "MPL-2.0"
  │            ^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: license is considered copyleft
  │
  = option-ext v0.2.0
    └── dirs-sys v0.4.1
        └── directories v5.0.1
            └── stackablectl v0.0.0-dev
                └── xtask v0.0.0-dev
```

sbernauer (approved):

Issue #98, opened 2023-08-18
Tracking: Release rewritten stackablectl officially using version `1.0.0`


Techassi:
```[tasklist]
### Build on supported platforms
- [ ] Build binary for Linux
- [ ] Build binary for MacOS
- [ ] Build binary for Windows
```

```[tasklist]
### Documentation
- [ ] Add and update documentation
- [ ] Auto generate parts of the documentation (help texts)
```

```[tasklist]
### Release Candidates
- [ ] Release `stackablectl-1.0.0-rc1` in the current state
- [ ] Release `stackablectl-1.0.0-rc2` with #94 (and maybe #85)
- [ ] Release `stackablectl-1.0.0-rc3` with #59
- [ ] Release `stackablectl-1.0.0` when documentation is up-to-date
```

```[tasklist]
### Misc
- [ ] Improve workflows
```

PR #106, opened 2023-08-24
fix(ctl): Skip creating operator-namespace when passing `--skip-release` in stack/demo command

Techassi:
Discovered by @sbernauer, brought forward in Slack chat.

### DIFF
--- a/rust/stackable-cockpit/src/platform/stack/spec.rs
+++ b/rust/stackable-cockpit/src/platform/stack/spec.rs
@@ -159,14 +159,8 @@ impl StackSpecV2 {
         release_list: ReleaseList,
         operator_namespace: &str,
         product_namespace: &str,
-        skip_release_install: bool,
     ) -> Result<(), StackError> {
         info!("Installing release for stack");
-
-        if skip_release_install {
-            info!("Skipping release installation during stack installation process");
-            return Ok(());
-        }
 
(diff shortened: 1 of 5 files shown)


rust/stackable-cockpit/src/platform/stack/spec.rs
```diff
@@ -159,15 +159,9 @@ impl StackSpecV2 {
         release_list: ReleaseList,
         operator_namespace: &str,
         product_namespace: &str,
-        skip_release_install: bool,
     ) -> Result<(), StackError> {
         info!("Installing release for stack");
```
sbernauer, line 163: Please remove the "for stack" (you can only install a release with no stack)" and mention the release name

sbernauer (dismissed):
LGTM % comment
In general the whole namespace creation is a bit scattered around (e.g. the release install function should ensure the operator namespaces exists, stacks and demos should not care about this) but should be fine ;)

Techassi:
Yes I agree. I already added a TODO on #98 to unify/refactor the namespace creation.

rust/stackable-cockpit/src/platform/stack/spec.rs
Techassi: Changed in baffb30

sbernauer (approved):

PR #111, opened 2023-09-07
ci: Improve workflow files

Techassi:
Tracked by #98 

This PR reworks and streamlines GitHub Action workflow files. It tries to reduce runs by selecting which workflows to run based on file paths. It also limits on which events these workflows get triggered.

We currently (and will never) allow direct pushes to `main`. That's why it is sufficient to only run on `pull_request` for most of the workflows. Release workflows on the other hand get triggered by pushes (to main) with tags attached.

**EDIT:** We might want to re-add the `merge_group` trigger for workflow files which currently only use `pull_request`.

- [x] Migrate from `actions-rs/clippy-check` to `giraffate/clippy-action`
- [x] Integrate old jobs from `Stackable Build Pipeline` into new workflows


.github/workflows/release_stackablectl.yml
```diff
@@ -48,17 +44,6 @@ jobs:
       - name: Build for non-Windows
         if: matrix.os != 'windows-latest'
         run: cargo build --target ${{ matrix.target }} --release -p stackablectl
-      - name: Prepare Windows env
```
sbernauer, line 51: For me this are valuable information once we try to get the Windows build to work.
You sure, you can get it from the git history, but you need to know *that* and *where* to look.

.github/workflows/release_stackablectl.yml
```diff
@@ -26,10 +26,6 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-20.04 # We don't use ubuntu-latest because we want to build against an old glibc version. (18.04 has glibc 2.27, 20.04 has glibc 2.31, 22.04 has glibc 2.35)
             file-suffix: ""
-          # Commented out, as Windows build is currently not working
```
sbernauer, line 29: same here

sbernauer (commented):
Did you have the chance to test it e.g. in a fork?

Techassi:
> Did you have the chance to test it e.g. in a fork?

No, but I can try it.

.github/workflows/release_stackablectl.yml
Techassi: Yeah I want to eventually replace this with the working build instructions for Windows.

.github/workflows/general_daily_security.yml
```diff
@@ -1,20 +1,16 @@
-# =============
```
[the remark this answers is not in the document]
Techassi: Yes. This is the ultimate goal. Depends how much we can re-use in the operator repos tho.

.github/workflows/pr_cockpit-web.yml
```diff
@@ -1,23 +1,18 @@
...
-  push:
-    branches:
-      - main
-      - staging
-      - trying
-      - "renovate/**"
-    tags:
-      - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
-  merge_group:
```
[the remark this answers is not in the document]
Techassi: If we check on every PR, we should not need it when merging as the code is tested and built beforehand. This ensures we don't push broken code into main.

.github/workflows/pr_cockpit-web.yml
```diff
@@ -1,23 +1,18 @@
...
+    paths:
+      - "package.json"
+      - "yarn.lock"
+      - "web"
```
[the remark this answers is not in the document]
Techassi: Agreed, will add the appropriate workflow files.

.github/workflows/pr_cockpit-web.yml
```diff
@@ -1,23 +1,18 @@
...
-    branches:
-      - main
-      - staging
-      - trying
-      - "renovate/**"
-    tags:
-      - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
-  merge_group:
+    paths:
```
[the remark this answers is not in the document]
Techassi: They just don't run at all and are not part of the required checks. They will be displayed as `skipped`.

.github/workflows/release_stackablectl.yml
Techassi: We will continue to not include the Windows build. I will also remove this commented out stuff, as it is not even close to what the final solution will look like (from my own experiments).

.github/workflows/release_stackablectl.yml
Techassi: See above.

.github/workflows/pr_cockpit-web.yml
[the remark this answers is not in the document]
Techassi: Added in 3822c0c

.github/workflows/release_stackablectl.yml
sbernauer: Alright!

sbernauer (approved):
I think we can start with this and improve it gradually if any problems arise

PR #118, opened 2023-09-27
docs: Update management pages

Techassi:
This PR adds more content to the management tools pages, with a focus on:

```[tasklist]
## Tasks
- [x] Adding documentation for subcommands (cache, completions, stack, stacklets)
- [x] What the statuses (Available, Reconciling, Running) mean
- [x] What a lack of a status means
- [x] Expand the Cockpit documentation
```


fhennig (changes requested):
Changes look generally good, thanks for expanding the docs! However a lot of the commandline options seem to be gone, is that correct? 

Also, I think it would be great if the generated docs would be generated int partials and then included from there. Then it would also be easier to make sure nobody changes them.

https://docs.antora.org/antora/latest/page/partials/

Techassi:
> However a lot of the commandline options seem to be gone, is that correct?

Yeah I was wondering the same. The code responsible for that didn't change. The output seems a little unstable.

> Also, I think it would be great if the generated docs would be generated int partials and then included from there. Then it would also be easier to make sure nobody changes them.

Very smart! I implemented that in 0235689

fhennig:
> Yeah I was wondering the same. The code responsible for that didn't change. The output seems a little unstable.

Well the way it is in the branch at the moment it can't be merged. It would be good to know what's going on :laughing: 

> Very smart! I implemented that in https://github.com/stackabletech/stackable-cockpit/commit/02356891f847b7c6b85dec3e7282941c0a785de7

Looks neat!

Techassi:
> Well the way it is in the branch at the moment it can't be merged. It would be good to know what's going on 😆

Yes, I will take a look.

fhennig (approved):

PR #156, opened 2023-12-04
fix: Semver version `pre` field doesn't contain dash

Techassi:
This fixes the helm repo selection based on the parsed version. The old code had several short-comings:

- The `pre` field doesn't include the dash used to separate version from pre. The matching however included this dash, resulting in wrong repo selection.
- Versions where `pre` starts with `pr` didn't probably select the Helm test repo.


NickLarsenNZ (approved):
Love the `rstest`s

siegfriedweber (approved):

PR #163, opened 2023-12-15
feat: allow plain output support

NickLarsenNZ:
# Description

Resolves #142 

- Rename old `plain` output option to `table`
- Ceate a new output option for `plain` which removes border characters and disables dynamic table arrangement.
  - There is an issue where there are multiple newlines between rows
  - There should be further discussion about what plain output should look like. This PR felt like a good-enough compromise between existing code and avoiding broken hyperlinks
  - We might need to use `force_no_tty()` when the output is plain, but I will look at that in a future PR where I add output options for `cache list`.

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes

```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Helm chart can be installed and deployed operator works
- [ ] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)

PR #164, opened 2023-12-15
fix: Update shell completions for previous PR (#163)

Techassi:
Small fixup for previous PR: It didn't include the updated shell completions.

I didn't realize these changes were missing in the PR. We should maybe add these checks to our workflows to assure we always merge an up2date version of these files.


NickLarsenNZ (approved):
Ah yeah, I knew there would be something.
Yes, we should automate that check

PR #174, opened 2024-01-03
fix: Uninstall operators when removing release

Techassi:
Fixes the release uninstall process by deleting installed operators correctly.

### DIFF
--- a/rust/stackable-cockpit/src/platform/release/spec.rs
+++ b/rust/stackable-cockpit/src/platform/release/spec.rs
@@ -8,7 +8,10 @@ use utoipa::ToSchema;
 
 use crate::{
     helm,
-    platform::{operator, product},
+    platform::{
+        operator::{self, OperatorSpec},
+        product,
+    },
 };
 
 type Result<T, E = Error> = std::result::Result<T, E>;
@@ -52,10 +55,10 @@ impl ReleaseSpec {
         info!("Installing release");
 
         for (product_name, product) in self.filter_products(include_products, exclude_products) {
-            info!("Installing product {}", product_name);
+            info!("Installing {}-operator", product_name);
 
             // Create operator spec
-            let operator = operator::OperatorSpec::new(product_name, Some(product.version.clone()))
+            let operator = OperatorSpec::new(product_name, Some(product.version.clone()))
                 .context(OperatorSpecParseSnafu)?;
 
             // Install operator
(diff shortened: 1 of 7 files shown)


rust/stackable-cockpit/src/platform/release/spec.rs
```diff
@@ -52,10 +55,10 @@ impl ReleaseSpec {
         info!("Installing release");
 
         for (product_name, product) in self.filter_products(include_products, exclude_products) {
-            info!("Installing product {}", product_name);
+            info!("Installing {}-operator", product_name);
```
sbernauer, line 58: Does the info!() macro expand the string *and* ship it as OpenTelemetry tag?
If not I would prefer the suggestion below
```suggestion
            info!("Installing {product_name}-operator");
```

rust/stackable-cockpit/src/platform/release/spec.rs
```diff
@@ -65,9 +68,20 @@ impl ReleaseSpec {
...
     }
 
+    #[instrument(skip_all)]
     pub fn uninstall(&self, namespace: &str) -> Result<()> {
-        for (product_name, _) in &self.products {
-            helm::uninstall_release(product_name, namespace, true).context(HelmUninstallSnafu)?;
+        info!("Uninstalling release");
+
+        for (product_name, product_spec) in &self.products {
+            info!("Uninstalling {}-operator", product_name);
```
sbernauer, line 76: Same as above

sbernauer (changes requested):
Thanks, LGTM in general

rust/stackable-cockpit/src/platform/release/spec.rs
Techassi: I'm actually not sure. I will read up on it!

Techassi: So basically this will be transformed into an event with a `message` field which uses the format string. Arguments specified after the format string will be used as arguments in said string. They will **not** be recorded as separate fields.

So we can use the suggested syntax.

rust/stackable-cockpit/src/platform/release/spec.rs
Techassi: Changed in 1cdab46

rust/stackable-cockpit/src/platform/release/spec.rs
Techassi: Changed in 1cdab46

sbernauer: Alright, many thanks for looking that up!

sbernauer (approved):

PR #181, opened 2024-01-09
fix: Use the provided cluster name when creating a local cluster

Techassi:
Fixes an error reported by @sbernauer on Slack.

The provided `--cluster-name` was not used when creating the local cluster either using `kind` or `minikube`.


sbernauer (approved):
Thanks a lot!

PR #204, opened 2024-02-12
feat: Add experimental debug container command

### LINKED ISSUE #201: Add command for creating a debug container

nightkr:
# Description

Adds a `stackablectl experimental-debug` command that creates and attaches a debug container. Compared to `kubectl debug` it should also (at least) mount in volumes and environment variables from the container being debugged.

Fixes #201.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


nightkr:
- [X] Create container
- [X] Attach
- [ ] Fall back to pulling logs if attach fails (maybe?)
- [x] Clean up stdin
- [x] Docs

nightkr:
This is intentionally _not_ part of the cockpit library since it's experimental and assumes that it's running in its own TTY. It could be factored out into a library component but I don't think it's super valuable to do preemptively.

NickLarsenNZ:
@nightkr, forgive my meddling. Was just trying to get the workflows to pass, so I added a couple of commits which were ultimately unnecessary as I backported some other changes to main which required those same two changes for the workflow to pass (I think). So, those two changes came in via the latest merge.

rust/stackablectl/Cargo.toml
```diff
@@ -29,5 +29,9 @@ serde.workspace = true
 snafu.workspace = true
 tera.workspace = true
 tokio.workspace = true
+tokio.features = ["io-std"]
```
Techassi: suggestion: Does it maybe make sense to use the object notation instead? It would be then inline with the root `Cargo.toml` file.

rust/stackablectl/Cargo.toml
```diff
@@ -29,5 +29,9 @@ serde.workspace = true
...
 tracing-subscriber.workspace = true
 tracing.workspace = true
+futures.workspace = true
+termion = "3.0.0"
+libc = "0.2.153"
```
Techassi: nitpick: For now we only need these dependencies for `stackablectl`, but does it maybe make sense to relocate these deps into the root `Cargo.toml` file?

rust/stackablectl/src/cmds/debug.rs
```diff
@@ -0,0 +1,380 @@
...
+    #[snafu(display("failed to update terminal size"))]
+    UpdateTerminalSize { source: mpsc::SendError },
+    #[snafu(display("container has no stdin channel"))]
+    NoStdinChannel,
+    #[snafu(display("container has no stdout channel"))]
+    NoStdoutChannel,
+    #[snafu(display("failed to forward stdin to container"))]
+    ForwardStdin { source: std::io::Error },
+    #[snafu(display("failed to forward stdout from container"))]
+    ForwardStdout { source: std::io::Error },
```
Techassi: note: Let's add a few newlines in between variants.

rust/stackablectl/src/cmds/debug.rs
```diff
@@ -0,0 +1,380 @@
...
+    ///
+    /// Volumes and environment variables will be copied from this container.
+    #[clap(long, short)]
+    container: String,
+    /// The debug container image
+    #[clap(long)]
+    image: String,
+    /// The command to run in the debug container
+    #[clap(last = true)]
+    cmd: Option<Vec<String>>,
```
Techassi: note: Also add newlines between fields.

rust/stackablectl/src/cmds/debug.rs
```diff
@@ -0,0 +1,380 @@
...
+            Ok(String::new())
+        }
+        .instrument(span)
+        .await
+    }
+}
+
+fn generate_debug_container_name() -> String {
+    let mut rng = rand::thread_rng();
+    let mut name = "sble-debug-".to_string();
```
Techassi: suggestion: Let's use the full name: `stackable-debug-abcde`.

rust/stackablectl/src/cmds/debug.rs
```diff
@@ -0,0 +1,380 @@
...
+    #[clap(long, short)]
+    namespace: Option<String>,
+    /// The Pod to debug
+    pod: String,
+    /// The target container to debug
+    ///
+    /// Volumes and environment variables will be copied from this container.
+    #[clap(long, short)]
+    container: String,
+    /// The debug container image
```
Techassi, line 123: suggestion: Does it maybe make sense to provide a default image suited for debugging? Or at least we could provide a suggestion for which image to use.

Because I'm already wondering which image you used while developing this feature.

rust/stackablectl/src/cmds/debug.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+/// Does true non-blocking reads of stdin, so that we can cancel properly on shutdown.
+/// The compromise is that it does not handle having things piped into it very well, since their write sides
+/// will also be turned non-blocking.
+///
+/// Only a single instance of AsyncStdin should exist at any one time.
+///
+/// Also, `AsyncStdin` will restore the mode of stdin when dropped.
+struct AsyncStdin {
```
Techassi, line 299: note: Aren't there any async Stdin implementations out there? If so, we could avoid maintaining our own implementation.

Techassi (changes requested):
First round, mostly looking good!

rust/stackablectl/Cargo.toml
nightkr: Or we can move it into the root Cargo.toml.

rust/stackablectl/Cargo.toml
nightkr: Good catch. I tend to mostly let `cargo add` do its thing, and then have a bad habit of not cleaning it up afterwards.

rust/stackablectl/Cargo.toml
Techassi: Sure, that's also possible!

rust/stackablectl/src/cmds/debug.rs
nightkr: Mostly alpine or archlinux. Honestly, this is a bit of an open question. I'm almost leaning towards the best default "for now" being the target container's image since it at least has the product tools installed.

nightkr: Should also look into running the debug container as root instead...

rust/stackablectl/src/cmds/debug.rs
nightkr: tokio's and async_std's both use a worker thread that calls the blocking APIs. That's better for compatibility with pipes, but leads to a dangling prompt when shutting down.

nightkr: Termion also has `AsyncReader` (https://docs.rs/termion/latest/termion/struct.AsyncReader.html) but that doesn't seem to have a wakeness trigger, so we don't know when we need to poll.

rust/stackablectl/src/cmds/debug.rs
Techassi: > Mostly alpine or archlinux. Honestly, this is a bit of an open question. I'm almost leaning towards the best default "for now" being the target container's image since it at least has the product tools installed.

Alright, then we should maybe add this as a hint to the arguments help text. It would definitely help to remove some of the guesswork involved when choosing a container image.

Regarding using a default value: I would leave it up to you.

rust/stackablectl/src/cmds/debug.rs
Techassi: Tricky topic! Then leave it as is I guess. Maybe in the future we can use some external implementation.

nightkr: I've tried reading through what OpenSSH does, and honestly it's kind of a mess that I couldn't really hope to untangle in a reasonable amount of time.. :/

rust/stackablectl/Cargo.toml
nightkr: 3fa1b7ee84a851a54c4b2bb4a5281ee52606cf0a

rust/stackablectl/Cargo.toml
nightkr: 3fa1b7ee84a851a54c4b2bb4a5281ee52606cf0a

rust/stackablectl/src/cmds/debug.rs
nightkr: 1567a8da411f86178862147014e7e1e14ca068a1

nightkr: 1567a8da411f86178862147014e7e1e14ca068a1

rust/stackablectl/src/cmds/debug.rs
nightkr: 1567a8da411f86178862147014e7e1e14ca068a1

rust/stackablectl/src/cmds/debug.rs
nightkr: 1567a8da411f86178862147014e7e1e14ca068a1

rust/stackablectl/src/cmds/debug.rs
nightkr: 335e1ac0c4fa14b68c223c487704b37bbde1f171

rust/stackablectl/src/cmds/debug.rs
nightkr: Default/docs: 9d26417dd1a7c1223b579d7099efefff8ebf0cbb
Root: bc4b32a6194c6082d249daf5439fd83628bbc2cd

Techassi (dismissed):

Techassi (approved):

PR #213, opened 2024-03-01
Support listing endpoints of Listeners in "stacklet list" command

sbernauer:
# Description

*Please add a description here. This will become the commit message of the merge request later.*

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/rust/stackable-cockpit/src/platform/service.rs
+++ b/rust/stackable-cockpit/src/platform/service.rs
@@ -45,15 +45,53 @@ pub async fn get_endpoints(
     object_name: &str,
     object_namespace: &str,
 ) -> Result<IndexMap<String, String>, Error> {
-    let service_list_params =
+    let mut endpoints = IndexMap::new();
+    let list_params =
         ListParams::from_product(product_name, Some(object_name), k8s::ProductLabel::Name);
 
-    let services = kube_client
-        .list_services(Some(object_namespace), &service_list_params)
+    let listeners = kube_client
+        .list_listeners(Some(object_namespace), &list_params)
         .await
         .context(KubeClientFetchSnafu)?;
 
-    let mut endpoints = IndexMap::new();
+    for listener in &listeners {
+        let Some(listener_status) = &listener.status else {
+            continue;
+        };
+        for address in listener_status.ingress_addresses.iter().flatten() {
+            for port in &address.ports {
+                // Listener name usually has pattern  listener-simple-hdfs-namenode-default-0 or
+                // simple-hdfs-datanode-default-0-listener, so we can strip everything before the first occurrence of
+                // the stacklet name (simple-hdfs in this case).
+                // This truncation is *not* ideal, however we only have implemented listener-operator for HDFS so far.
+                let listener_name = listener.name_any();
+                let Some((_, display_name)) = listener_name.split_once(object_name) else {
+                    continue;
+                };
+                let text = format!(
+                    "{display_name}-{port_name}",
+                    display_name = display_name.trim_start_matches('-'),
+                    port_name = port.0
+                );
+                let endpoint_url = endpoint_url(&address.address, *port.1, port.0);
+                endpoints.insert(text, endpoint_url);
+            }
+        }
+    }
+
+    // Ideally we use listener-operator everywhere, afterwards we can remove the whole k8s Services fetching.
+    // Currently the Services created by listener-op are missing the recommended labels, so early exit in case we find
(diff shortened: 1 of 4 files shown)


rust/stackable-cockpit/src/platform/service.rs
```diff
@@ -45,15 +45,54 @@ pub async fn get_endpoints(
     object_name: &str,
     object_namespace: &str,
 ) -> Result<IndexMap<String, String>, Error> {
-    let service_list_params =
+    let mut endpoints = IndexMap::new();
```
Techassi: note: Why do we create a `IndexMap` here, even if we could fail at line 55. I would suggest to move the instantiation back below the K8s client list call.

rust/stackable-cockpit/src/platform/service.rs
```diff
@@ -45,15 +45,54 @@ pub async fn get_endpoints(
...
         .context(KubeClientFetchSnafu)?;
 
-    let mut endpoints = IndexMap::new();
+    for listener in &listeners {
+        let Some(listener_status) = &listener.status else {
+            continue;
+        };
+        for address in listener_status.ingress_addresses.iter().flatten() {
+            for port in &address.ports {
+                // Listener names usually have the pattern "listener-simple-hdfs-namenode-default-0" or
```
Techassi: suggestion: We should make the naming consistent. Then we could add a stable function which extracts the raw display name.

rust/stackable-cockpit/src/platform/service.rs
```diff
@@ -45,15 +45,54 @@ pub async fn get_endpoints(
...
+                endpoints.insert(text, endpoint_url);
+            }
+        }
+    }
+
+    // Ideally we use listener-operator everywhere, afterwards we can remove the whole k8s Services handling following.
+    // Currently the Services created by listener-op are missing the recommended labels, so this early exit in case we
+    // find Listeners is currently not required. However, once we add the recommended labels to the k8s Services, we
+    // would have duplicated entries (one from the Listener and one from the Service). Because of this we don't look at
+    // the Services in case we found Listeners!
```
Techassi: suggestion: Slight text adjustment

```suggestion
    // Ideally we use listener-operator everywhere, afterwards we can remove the whole k8s Services handling below.
    // Currently the Services created by listener-op are missing the recommended labels, so this early exit in case we
    // find Listeners is currently not required. However, once we add the recommended labels to the k8s Services, we
    // would have duplicated entries (one from the Listener and one from the Service). Because of this we don't look at
    // the Services in case we found Listeners!
```

Techassi (changes requested):

rust/stackable-cockpit/src/platform/service.rs
sbernauer: No real reason, changed in 4c7b9df

rust/stackable-cockpit/src/platform/service.rs
sbernauer: The problem is that the ListenerClass name is equal to the name of the PVC that creates it.
And the pvc naming is done by k8s based on the name of the Pod and the name of the volume in the PodSpec.
And it also differs depend on wether a persistent or ephemeral volume is used :/ So everything is a mess....

Techassi: Well, that's just... great. I guess we can make it stable(ish) sometime in the future. Is it still worth pulling the code out into its own function?

Techassi (approved):

rust/stackable-cockpit/src/platform/service.rs
sbernauer: Sure, please have a look at 1a11e41

Techassi: Nice change, I like the addition of some test cases!

Techassi (approved):

lfrancke:
What does this ticket fix? Only HDFS? Everything?

rust/stackable-cockpit/src/platform/service.rs
nightkr: Oh, huh. Guess I missed sending my reply here the other day (or GH ate it). I'd personally much rather just show the name as-is than mangle it like this. That would also make it easier/clearer for users to move between the "easy facade" of stackablectl and reading the backend data via kubectl when something goes wrong.

PR #226, opened 2024-03-21
chore: Release stackablectl-24.3.1

Techassi:
Includes the experimental `debug` command.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2764,7 +2764,7 @@ dependencies = [
 
 [[package]]
 name = "stackablectl"
-version = "24.3.0"
+version = "24.3.1"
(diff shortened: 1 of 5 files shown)


Techassi:
>  but I really don't like us versioning stackablectl in lockstep with the platform and than doing a bugfix release for only parts of it. I fear this tricks users into thinking we have shipped a 24.3.1 platform release.

That's why I initially voted for a separate version, like `1.0.0`.

We might eventually want to do it the same way Kubernetes does it with `kubectl`. It is versioned independently and offers a support matrix. We could achieve the same thing by having long running clusters in different SDP versions against which we run various `stackablectl` commands to verify it works.

PR #233, opened 2024-04-24
chore: Bump Rust and Node deps

Techassi:
This bumps various Rust and Node dependencies. Currently `kube` cannot be bumped, because `stackable-operator` still uses `0.88.1`. This can be done once a new version of `stackable-operator` is releases.

Closes #231, closes #216. Other Node deps can be bumped as well, but might require adjustments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,10 @@ repository = "https://github.com/stackabletech/stackable-cockpit/"
 
 [workspace.dependencies]
 async-trait = "0.1"
-axum = { version = "0.6", features = ["http2", "headers"] }
+axum = { version = "0.7", features = ["http2"] }
+axum-extra = { version = "0.9", features = ["typed-header"] }
 bcrypt = "0.15"
-bindgen = "0.68.1"
+bindgen = "0.69.4"
 cc = "1.0.83"
 clap = { version = "4.2.1", features = ["derive", "env"] }
 clap_complete = "4.2"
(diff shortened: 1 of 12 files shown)


Techassi:
@sbernauer let me know if I should bump various other Node deps (which might require more work) in this PR.

Cargo.toml
```diff
@@ -27,10 +28,10 @@ directories = "5.0"
 dotenvy = "0.15"
 futures = "0.3"
 indexmap = { version = "2.0", features = ["serde"] }
-k8s-openapi = { version = "0.20", default-features = false, features = [
+k8s-openapi = { version = "0.21", default-features = false, features = [
```
sbernauer, line 31: FYI https://github.com/stackabletech/operator-rs/pull/769. Not saying we should block this though

rust/stackablectl/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Add pre-built binary for `aarch64-unknown-linux-gnu` ([#232]).
+
+[#232]: https://github.com/stackabletech/stackable-cockpit/pull/232
```
sbernauer: Nit: Normally we group all links for a link at the bottom (instead of having one list for added and one for changed

sbernauer (approved):
Thanks, only nits. I think I would prefer a separate PR, as I'm not familiar with Node at all

rust/stackablectl/CHANGELOG.md
Techassi: Ah yeah... I will fix that. Does it maybe make sense to split them up? Especially when we have to deal with an increasing amount of entries in the changelog?

Cargo.toml
Techassi: Yeah we can wait for the `stackable-operator` release to bump `k8s-openapi` and friends.

rust/stackablectl/CHANGELOG.md
Techassi: Changed in 834a91a

sbernauer (approved):

sbernauer (approved):

PR #238, opened 2024-05-13
chore: Bump Rust, Go and Node dependencies

Techassi:
# Description

This PR bumps various Rust and Node dependencies. Should resolve #185, #211, #221, #225, #229 and #230.

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

PR #239, opened 2024-05-13
chore: Release stackablectl-24.3.3

Techassi:
# Description

Releases `stackablectl-24.3.3` to include the bumped dependencies from #238.

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


sbernauer (approved):
Here we go again. Still don't like the versioning xD

Techassi:
> Still don't like the versioning xD

Yes, I'm not happy about it either. I used to use `1.X.X`, but we voted against using it in the past, so let's create a decision for that (I can write something up).

We could aim to use a similar versioning scheme `kubectl` uses and provide a support matrix.

sbernauer (dismissed):

sbernauer (approved):
Even better

PR #300, opened 2024-06-17
fix: Remove colons from error messages

Techassi:
The snafu report cleans up error messages by removing any content after a colon is encountered in the error message. This also messes up the output of structs when the Debug output is used. This was causing an issue when we failed to resolve a GVK.


NickLarsenNZ (approved):
LGTM

PR #301, opened 2024-06-17
chore: Release stackablectl-24.3.5

Techassi:
This release includes:

### Fixed

- Remove colons from error messages, because the snafu report removes all content after the colon which results in loss of detail ([#300]).

[#300]: https://github.com/stackabletech/stackable-cockpit/pull/300


NickLarsenNZ (approved):
LGTM

PR #302, opened 2024-06-17
fix: update Cargo.nix version and pre-commit hook improvements

NickLarsenNZ:
> [!IMPORTANT]
> We probably want to retag `24.3.5` once this is merged.

# Description

Without Cargo.nix being updated (via make regenerate-nix), the old version was still appearing in the new binary.

Additional improvements to speed up the pre-commit hook feedback loop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,11 +6,6 @@
 htpasswd
 .yarn
 
-# Nix
-gomod2nix.toml
-Cargo.nix
-crate-hashes.json
-
 # Misc
 /demos
 /stacks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,30 +50,35 @@ repos:
     hooks:
       - id: gen-man
         name: gen-man
+        files: ^rust/stackablectl/
         language: system
         entry: cargo xtask gen-man
         stages: [commit, merge-commit, manual]
         pass_filenames: false
       - id: gen-comp
         name: gen-comp
+        files: ^rust/stackablectl/
         language: system
         entry: cargo xtask gen-comp
         stages: [commit, merge-commit, manual]
         pass_filenames: false
       - id: gen-openapi
         name: gen-openapi
+        files: ^web/
         language: system
         entry: cargo xtask gen-openapi
         stages: [commit, merge-commit, manual]
         pass_filenames: false
       - id: gen-ctl-readme
         name: gen-ctl-readme
+        files: ^rust/stackablectl/
         language: system
         entry: cargo xtask gen-ctl-readme
         stages: [commit, merge-commit, manual]
         pass_filenames: false
       - id: gen-docs
         name: gen-docs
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -6,11 +6,6 @@
 htpasswd
 .yarn
 
-# Nix
-gomod2nix.toml
```
Techassi, line 10: suggestion: Add the `gomod2nix.toml` file to the `.gitattributes` file to mark it as auto-generated.

.pre-commit-config.yaml
```diff
@@ -1,14 +1,16 @@
...
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.2.0
     hooks:
       - id: trailing-whitespace
         # Generated file
-        exclude: ^web/src/api/schema.d.ts|extra/.*$
+        exclude: ^web/src/api/schema\.d\.ts|extra/.*$
       - id: end-of-file-fixer
+        exclude: Cargo.nix
```
Techassi, line 13: note: Ideally, this would be fixed upstream, but for now this is fine.

web/src/api/schema.d.ts
```diff
@@ -142,8 +142,6 @@ export interface components {
   pathItems: never;
 }
 
-export type $defs = Record<string, never>;
```
Techassi: note: This is rather weird. It seems like people use slightly different versions (even tho it is specified in `package.json` and `yarn.lock`) which produces a different output each time.

Techassi (commented):
LGTM, just a bunch of questions / notes.

.pre-commit-config.yaml
NickLarsenNZ: Upstream maintainers might not see it as something needing to be fixed (some people don't care about trailing new lines), so I think this is not only _for now_.

What would be nice if the hook could ignore generated files, eg: what we have marked as `linguist-generated` in `.gitattributes` so we didn't have to maintain two sets of rules.

.gitignore
NickLarsenNZ: Oh, for some reason I thought that some some config (just because it was toml)...
Done in e2a4d70bbd67375ff57f21ec40412bf7d233152a

web/src/api/schema.d.ts
NickLarsenNZ: Turns out I needed to `yarn install`, oops.

Reverted in c8f8aaac7a4c89236e5d5f1198b720be8fcb1b72

.pre-commit-config.yaml
Techassi: > What would be nice if the hook could ignore generated files, eg: what we have marked as linguist-generated in .gitattributes so we didn't have to maintain two sets of rules.

That would be really nice indeed.

Techassi (approved):

PR #306, opened 2024-06-25
fix(helm-sys): Double the helm timeout after reported timeout errors

NickLarsenNZ:
# Description

There have been multiple reports of errors when deploying to some clusters.

image

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/rust/helm-sys/go-helm-wrapper/main.go
+++ b/rust/helm-sys/go-helm-wrapper/main.go
@@ -39,7 +39,7 @@ func main() {
 func go_install_helm_release(releaseName *C.char, chartName *C.char, chartVersion *C.char, valuesYaml *C.char, namespace *C.char, suppressOutput bool) *C.char {
 	helmClient := getHelmClient(namespace, suppressOutput)
 
-	timeout, _ := time.ParseDuration("10m")
+	timeout, _ := time.ParseDuration("20m")
 	chartSpec := gohelm.ChartSpec{
 		ReleaseName: C.GoString(releaseName),
 		ChartName:   C.GoString(chartName),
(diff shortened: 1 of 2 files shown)


rust/helm-sys/go-helm-wrapper/main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 func go_install_helm_release(releaseName *C.char, chartName *C.char, chartVersion *C.char, valuesYaml *C.char, namespace *C.char, suppressOutput bool) *C.char {
 	helmClient := getHelmClient(namespace, suppressOutput)
 
-	timeout, _ := time.ParseDuration("10m")
+	timeout, _ := time.ParseDuration("20m")
```
razvan, line 42: I'm not a go expert but I think this is the wrong timeout to change. Judging by the error message, I believe the context passed to `InstallChart() ` is the proper one to update.

Techassi: The timeout passed to the `ChartSpec`

> configures the time to wait for any individual Kubernetes operation (like Jobs for hooks).

_(per documentation, see [here](https://pkg.go.dev/github.com/mittwald/go-helm-client#ChartSpec))_

The `context.Background()` function returns a context, which will never cancel. We could pass a timeout here as well, but that would configure the timeout for the _complete_ operation, not individual K8s operations.

So I think the change looks fine in my opinion.

razvan (approved):
I guess it's worth a try.

NickLarsenNZ:
> I guess it's worth a try.

That's the spirit :smile:

PR #307, opened 2024-07-11
chore: Bump Rust and Go dependencies

NickLarsenNZ:
# Description

*Please add a description here. This will become the commit message of the merge request later.*

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


rust/stackablectl/src/cmds/debug.rs
```diff
@@ -353,6 +353,7 @@ impl AsyncRead for AsyncStdin {
     ) -> Poll<std::io::Result<()>> {
         loop {
             let mut ready = ready!(self.fd.poll_read_ready_mut(cx)?);
+            #[allow(clippy::blocks_in_conditions)]
```
NickLarsenNZ: @Techassi I couldn't find a way to resolve this lint without it being arguably more complex with the addition of type definitions for the `r`.

rust/stackable-cockpitd/src/api_doc.rs
```diff
@@ -57,7 +57,7 @@ pub fn openapi() -> utoipa::openapi::OpenApi {
 }
 
 /// Synthetic types that are used to generate type definitions for foreign types.
-mod synthetic_types {
+pub mod synthetic_types {
```
NickLarsenNZ: I assume this is ok. Otherwise we can do:

```suggestion
#[allow(dead_code)]
mod synthetic_types {
```

rust/stackablectl/src/cmds/debug.rs
Techassi: I will take a look.

Techassi: This seems to work:

```rust
loop {
    let mut ready = ready!(self.fd.poll_read_ready_mut(cx)?);

    let io = ready.try_io(|r| {
        let read = r.get_mut().read(buf.initialize_unfilled())?;
        buf.advance(read);
        Ok(())
    });

    break match io {
        Ok(res) => Poll::Ready(res),
        Err(_would_block) => {
            // Try to poll again, so that we re-register the waker
            continue;
        }
    };
}
```

NickLarsenNZ: Thanks @Techassi, done in 85214f495b36cf42be67966518391b3e8ebcd820

rust/stackable-cockpitd/src/api_doc.rs
Techassi: This _should_ be fine, as this is not a library and as such doesn't expose any of the types (and modules in this case).

I still think that this should be kept private, because it is only used during OpenAPI spec generation (and I thus consider it as an internal implementation detail). So I would prefer to add `#[allow(dead_code)]` here. But let's also add a comment for what the module is used and why we added the attribute.

Techassi:
Hint: The `cargo udeps` error is caused by an outdated `cargo-udeps` version. Bumping the version used should fix the issue, see https://github.com/stackabletech/operator-rs/pull/804 for additional details.

rust/stackable-cockpitd/src/api_doc.rs
NickLarsenNZ: There is already a comment.
I'll keep it private, and use the `allow` instead.

Techassi: Could we maybe extend the comment and state why the attribute was needed? For the future reader it might be immediately obvious why it's there.

NickLarsenNZ: Done in 577b4534e9b48b81db1918b9c31ebe894fd79183

NickLarsenNZ:
> Hint: The `cargo udeps` error is caused by an outdated `cargo-udeps` version. Bumping the version used should fix the issue, see [stackabletech/operator-rs#804](https://github.com/stackabletech/operator-rs/pull/804) for additional details.

Thanks. I'll update it, but first I want to see what happens if I only update the rust-toolchain in CI. ~I have a sneaking suspicion that it will _just work_.~ It didnt't work :sweat_smile:

NickLarsenNZ:
Should I include go updates in this too?
And is there anything else to consider ahead of the release?

Techassi:
> Should I include go updates in this too?

Yes, we can update those as well. They might get a little tricky due to the FFI/cgo stuff.

> And is there anything else to consider ahead of the release?

Nothing I can think of.

NickLarsenNZ (commented):
LGTM

I have tested installing a stack and listing the stacklets. :rocket:

Techassi (approved):

PR #311, opened 2024-07-26
Restore broken Nix build

nightkr:
# Description

24.7.0 breaks the Nix build for a few reasons:

1. utoipa-swagger-ui now supports file:// URLs (https://github.com/juhaku/utoipa/pull/923), so we need to provide a real file path
2. We now require Go >= 1.22.5, which is not available in the pinned Nixpkgs

This PR fixes 1 directly, and 2 by updating all pinned Nix libraries.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -24,20 +24,11 @@
       };
       utoipa-swagger-ui = attrs: rec {
         # utoipa-swagger-ui tries to redownload swagger-ui, which is blocked by Nix's sandboxing
-        # so we download it instead, and put it where it expects to be cached
-        preConfigure =
-          let
-            swaggerUi = pkgs.fetchurl {
-              url = "https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.3.zip";
-              hash = "sha256-zrb8feuuDzt/g6y7Tucfh+Y2BWZov0soyNPR5LBqKx4=";
-            };
-          in ''
-            mkdir -p target/build/utoipa-swagger-ui.out
-            ln -s "${swaggerUi}" target/build/utoipa-swagger-ui.out/swagger-ui.zip
-          '';
-        # Build script only supports fetching from HTTP, not file URLs
-        # Last path element decides the cache key, which we rely on above
-        SWAGGER_UI_DOWNLOAD_URL = "file:///invalid-path/swagger-ui.zip";
+        # so we download it instead, and tell it to use that
+        SWAGGER_UI_DOWNLOAD_URL = "file://${pkgs.fetchurl {
+          url = "https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.3.zip";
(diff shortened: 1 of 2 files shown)


NickLarsenNZ (approved):
I assume you just did a `niv update`, but if it was something else, would you mind noting it in the description (just so it is stored in the merge commit).

nightkr:
Yep, 2 was just a niv update.

default.nix
```diff
@@ -24,20 +24,11 @@
...
-            ln -s "${swaggerUi}" target/build/utoipa-swagger-ui.out/swagger-ui.zip
-          '';
-        # Build script only supports fetching from HTTP, not file URLs
-        # Last path element decides the cache key, which we rely on above
-        SWAGGER_UI_DOWNLOAD_URL = "file:///invalid-path/swagger-ui.zip";
+        # so we download it instead, and tell it to use that
+        SWAGGER_UI_DOWNLOAD_URL = "file://${pkgs.fetchurl {
+          url = "https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.3.zip";
+          hash = "sha256-zrb8feuuDzt/g6y7Tucfh+Y2BWZov0soyNPR5LBqKx4=";
+        }}";
```
brhutchins, line 31: :chefkiss:

default.nix
```diff
@@ -24,20 +24,11 @@
...
-              hash = "sha256-zrb8feuuDzt/g6y7Tucfh+Y2BWZov0soyNPR5LBqKx4=";
-            };
-          in ''
-            mkdir -p target/build/utoipa-swagger-ui.out
-            ln -s "${swaggerUi}" target/build/utoipa-swagger-ui.out/swagger-ui.zip
-          '';
-        # Build script only supports fetching from HTTP, not file URLs
-        # Last path element decides the cache key, which we rely on above
-        SWAGGER_UI_DOWNLOAD_URL = "file:///invalid-path/swagger-ui.zip";
+        # so we download it instead, and tell it to use that
```
dehfrancis, line 27: @nightkr this happens to solve my problem as well. But i am currently facing a situation where nix throws this error when trying to copy the zip file to the target build folder.
```
{ code: 13, kind: PermissionDenied, message: "Permission denied" }
```
Is this something you have faced before? i'd appreciate your thoughts on this.

nightkr: As in the Nix daemon itself throws the error, or build.rs does? Can't say it's something I recognize immediately, but if I were you I'd try to patch utoipa-swagger-ui's build.rs to dump out the environment (what absolute path is it writing to, who's it running as, how much of the path exists, what're the permissions like, etc etc etc).

dehfrancis: yes the build.rs does throws this during the nix build. and yes i can identify the various environment related details. Except i'm new to nix and i don't know if i actually can modify the permissions so the build passes. but i'll research more on it. thanks for your prompt response.

PR #320, opened 2024-09-05
fix(ci): Install hadolint in pre-commit workflow

Techassi:
Part of https://github.com/stackabletech/issues/issues/637

This fixes two things:

- Install the hadolint binary to be able to run the pre-commit hook
- Set the default language version of `node` to `system`. This fixes issues on NixOS systems.

### DIFF
--- a/.github/workflows/pr_pre-commit.yml
+++ b/.github/workflows/pr_pre-commit.yml
@@ -22,6 +22,15 @@ jobs:
         with:
           toolchain: ${{ env.RUST_TOOLCHAIN_VERSION }}
           components: rustfmt,clippy
+      - name: Setup Hadolint
+        shell: bash
+        run: |
+          LOCATION="$HOME/.local/bin/hadolint"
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_pre-commit.yml
```diff
@@ -18,10 +19,22 @@ jobs:
       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: '3.12'
-      - uses: dtolnay/rust-toolchain@master
```
NickLarsenNZ: Is this not required? I thought it was, for clippy, etc...

Techassi: Yeah, it is required. I accidentally removed it.

Techassi: Re-added in 75fed3e

NickLarsenNZ (approved):
LGTM

PR #339, opened 2024-12-03
ci: Update run-pre-commit action

Techassi:
> [!CAUTION]
> This can only be merged when a tag was added in https://github.com/stackabletech/actions

This now installs Nix via the run-pre-commit action instead.


NickLarsenNZ (approved):
LGTM

Issue #310, opened 2024-07-25
feat: Add ability to select version of demos/stacks


NickLarsenNZ:
~Once demos are versioned per release (see sub-issue)~ DONE, we should be able to select which demo/stack version to install.

Suggested option to add:

```shell
stackablectl demo in monitoring --release 24.7
```

The default should be the [latest](https://github.com/stackabletech/release/blob/75bab4847b2c2684aa379ff7f82ff9a2a6e79533/releases.yaml#L308-L311) stable release.

Examples (competing ideas, in no particular order)

```sh
stackablectl demo in monitoring # latest stable. Currently this deploys manifests from `main`

# Enforce a particular release
stackablectl demo in monitoring --release 24.7
stackablectl demo in monitoring --release dev

# Take any committish (stackableRelease derived from stack.yaml at that version)
stackablectl demo in monitoring --ref abcdef7
stackablectl demo in monitoring --ref release-24.7

# Or, enforce a branch (because the HEAD is what should be working. Random commits might cause surprises)
stackablectl demo in monitoring --branch release-24.7 # but must be a branch
```

Considerations:

- `--ref` ([Committish]), or `--branch` (and check that it is a branch), or `--release` (we need to derive the ref).
  - We can work out the operator release, because at that version
  - We should ideally go with the most flexible (`--ref`), unless we cannot derive the release.
  - ⚠ There will still be hard coded refs in the stack/demo yamls, as well as scripts. This can lead to broken demos.
  - Specifying `--release` would remove the need for the `stackableRelease` key in `stacks.yaml`.
- Caching - do we need to invalidate the cache if a ref/branch/release is specified?
- Overlap with: https://github.com/stackabletech/stackable-cockpit/issues/212

[Committish]: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish

NickLarsenNZ:
Not having this is making things really difficult at release time, because the current behaviour:
- requires the demos `main` branch to be stable, which means...
- changes needed for an upcoming release need to be held back in a `next` branch (which still can't be used by stackablectl without checking out the branch), and then...
    - The `next` branch gets mixed with changes that can work for the next release, as well as other changes that break it (eg: image tags that don't yet exist). This confusion could be avoided by using the `0.0.0-dev` tags.
- Near release time, the `next` branch gets merged - generally breaking the stability promise of the `main` branch.

Notes:
[...]

PR #340, opened 2024-12-05
feat: Select versions of demos & stacks

labrenbe:
# Description

Fixes https://github.com/stackabletech/stackable-cockpit/issues/310.

- Add the `--release <release>` argument to `demo` and `stack` commands which allows installing from a specific release branch.
- If `--release` is not specified the latest release is used.
- `--release dev` install from main branch.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


rust/stackablectl/src/cli/mod.rs
```diff
@@ -92,25 +87,30 @@ impl Cli {
     /// Returns a list of demo files, consisting of entries which are either a path or URL. The list of files combines
     /// the default demo file URL, [`REMOTE_DEMO_FILE`], files provided by the ENV variable [`ENV_KEY_DEMO_FILES`], and
     /// lastly, files provided by the CLI argument `--demo-file`.
-    pub fn get_demo_files(&self) -> Result<Vec<PathOrUrl>, PathOrUrlParseError> {
-        let mut files = get_files(REMOTE_DEMO_FILE, ENV_KEY_DEMO_FILES)?;
+    pub fn get_demo_files(&self, branch: &str) -> Result<Vec<PathOrUrl>, PathOrUrlParseError> {
```
Techassi, line 90: note: The function should now mention the new `branch` parameter it accepts and what it does.

rust/stackablectl/src/cli/mod.rs
```diff
@@ -92,25 +87,30 @@ impl Cli {
...
-        let files = self.get_demo_files().unwrap();
-        List::build(&files, transfer_client).await.unwrap()
-    }
-
     /// Returns a list of stack files, consisting of entries which are either a path or URL. The list of files combines
     /// the default stack file URL, [`REMOTE_STACK_FILE`], files provided by the ENV variable [`ENV_KEY_STACK_FILES`],
     /// and lastly, files provided by the CLI argument `--stack-file`.
-    pub fn get_stack_files(&self) -> Result<Vec<PathOrUrl>, PathOrUrlParseError> {
-        let mut files = get_files(REMOTE_STACK_FILE, ENV_KEY_STACK_FILES)?;
+    pub fn get_stack_files(&self, branch: &str) -> Result<Vec<PathOrUrl>, PathOrUrlParseError> {
```
Techassi, line 108: note: Same as above, mention the new parameter.

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -278,6 +316,7 @@ async fn install_cmd(
     cli: &Cli,
     list: demo::List,
     transfer_client: &xfer::Client,
+    release_branch: &str,
 ) -> Result<String, CmdError> {
     info!("Installing demo {}", args.demo_name);
```
Techassi: suggestion: We should add the release branch as a field here.

```suggestion
    info!(%release_branch, "Installing demo {}", args.demo_name);
```

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -155,18 +165,46 @@ impl DemoArgs {
 
         let transfer_client = xfer::Client::new_with(cache);
 
+        let release_files = cli.get_release_files().context(PathOrUrlParseSnafu)?;
+        let release_list = release::ReleaseList::build(&release_files, &transfer_client)
+            .await
+            .context(BuildListSnafu)?;
+
+        let release_branch = match &self.release {
```
Techassi, line 173: note: This block of could looks like it can be added as a function to the `ReleaseList` type. This way, we can save duplicate code.

Techassi: note: We might also want to confirm that the release which the user provided via `--release` actually exists.

Something similar to `release_list.get(release)` comes to mind: `ReleaseList::contains(release) -> bool` for example. If this function returns false, we could use the (currently unused) `NoSuchRelease` error variant.

rust/stackablectl/src/constants.rs
```diff
@@ -2,6 +2,8 @@ pub const ENV_KEY_RELEASE_FILES: &str = "STACKABLE_RELEASE_FILES";
 pub const ENV_KEY_STACK_FILES: &str = "STACKABLE_STACK_FILES";
 pub const ENV_KEY_DEMO_FILES: &str = "STACKABLE_DEMO_FILES";
 
+pub const DEMOS_REPOSITORY_URL_BASE: &str = "https://raw.githubusercontent.com/stackabletech/demos";
```
Techassi, line 5: question: Are the other constants (like `REMOTE_DEMO_FILE` and similar) used somewhere else. If so, it might cause problems based on the changes this PR introduces.

Techassi: note: It might be good to add at least two new constants, containing `demos/demos-v2.yaml` and `stacks/stacks-v2.yaml` and using them constructing the URL.

Techassi (changes requested):
Mostly lgtm, a few smaller comments.

Also, please add changelog entries regarding the `--offline` removal, the addition of `--release` and make sure to fix pre-commit and rust-doc errors.

rust/stackablectl/src/cmds/stack.rs
```diff
@@ -140,7 +147,34 @@ impl StackArgs {
...
+                if release == "dev" {
+                    "main".to_string()
+                } else {
+                    format!("release-{release}")
+                }
+            }
+            None => {
+                let release = release_list
+                    .inner()
+                    .first()
```
sbernauer: I think we can not just always take the first item from the list (there can also be user-provided releases).
We can also probably also not sort them by semver because of that.
But we can probably avoid all of that by using `latest`

rust/stackablectl/src/constants.rs
labrenbe: `REMOTE_DEMO_FILE` and `REMOTE_STACK_FILE` are not used elsewhere in this or other repos and should will removed.

rust/stackablectl/src/cli/mod.rs
Techassi: Done in e6f73ca.

rust/stackablectl/src/cli/mod.rs
Techassi: Done in e6f73ca.

rust/stackablectl/src/constants.rs
Techassi: Done in e6f73ca.

rust/stackablectl/src/cmds/demo.rs
Techassi: Done in 01915ed.

Techassi (approved):

PR #337, opened 2024-11-19
feat: Add shell completions for Nushell and Elvish

NickLarsenNZ:
# Description

- Nushell completions.
  _Uses the [clap_complete_nushell] crate, as linked by the already used [clap_complete] crate._
- Elvish completions.
  _Already included in [clap_complete], so why not?_ 
- Update xtasks to generate the new shell completions.
- New xtask command to run all xtasks (useful for local development).
- Disable shellcheck on generated completions.

[clap_complete_nushell]: https://crates.io/crates/clap_complete_nushell
[clap_complete]: https://crates.io/crates/clap_complete

PR #345, opened 2025-01-13
docs: Add new completions, simplify install instructions

Techassi:
This PR adds instructions for Nushell and Elvish completions (added in #337). It additionally simplifies the installation instructions.

**Preview:**

image


NickLarsenNZ (approved):
LGTM.

In a future PR we can switch the shell-completion stuff around to use the "current stackablectl" binary instead if pulling potentially conflicting completions.

Techassi:
Yes, that sounds reasonable!

PR #351, opened 2025-01-31
fix: Use native CA certificates

NickLarsenNZ:
# Description

Use `rustls-native-certs` so that `stackablectl` can be used in environments with internal PKI.

This is needed in the following cases:
- Internal mirrors of helm charts or stacks/demos signed by an internal CA.
- Firewalls that intercept HTTPS traffic.

### DIFF
--- a/rust/stackablectl/Cargo.toml
+++ b/rust/stackablectl/Cargo.toml
@@ -2,7 +2,7 @@
 name = "stackablectl"
 description = "Command line tool to interact with the Stackable Data Platform"
 # See <project-root>/Cargo.toml
-version = "24.11.3"
+version = "24.11.4-native-certs"
 authors.workspace = true
 license.workspace = true
 edition.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2500,6 +2500,7 @@ dependencies = [
  "pin-project-lite",
  "quinn",
  "rustls",
+ "rustls-native-certs",
  "rustls-pemfile",
  "rustls-pki-types",
  "serde",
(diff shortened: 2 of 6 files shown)


rust/stackablectl/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "stackablectl"
 description = "Command line tool to interact with the Stackable Data Platform"
 # See <project-root>/Cargo.toml
-version = "24.11.3"
```
Techassi, line 5: question: Do we want this to be in `main`? I guess you could argue in both directions.

Techassi:
The change itself is approved. We just need to clarify the above question.

rust/stackablectl/Cargo.toml
NickLarsenNZ: I'd prefer it to be `0.0.0-dev` like other things. Then it gets bumped. But that would require release branches to put the update on.

Techassi: Yeah I don't see the need for release branches in this repo. It would like to keep it as is which is the same as many other (Rust-based) tools are doing it.

NickLarsenNZ: The downside to that is:
- the version in `main` does not represent the actual thing at that version.
- building release-candidates on branches requires manually updating the version (as I have done in this PR), which then needs changing back (which I have forgotten to do), or pre-empting.

I'll leave it as is and it can be corrected in the next PR to release stackablectl.

Techassi: We can leave it as is for now (and change it back to a "normal" version once we release again).

It might be worth to invest some time to adapt the release workflow to also handle release candidates and mark the GitHub releases as a draft.

Techassi (approved):

PR #355, opened 2025-02-26
feat: Support templating the namespace in manifests

sbernauer:
# Description

Fixes https://github.com/stackabletech/stackable-cockpit/issues/371
Also a bit of clean-up from https://github.com/stackabletech/stackable-cockpit/pull/373

After this feature we should actually migrate demos to don't hard-code `default`.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -74,6 +74,11 @@ pub trait InstallManifestsExt {
     ) -> Result<(), Error> {
         debug!("Installing demo / stack manifests");
 
+        let mut parameters = parameters.clone();
+        // We add the NAMESPACE parameter, so that stacks/demos can use that to render e.g. the
+        // fqdn service names [which contain the namespace].
+        parameters.insert("NAMESPACE".to_owned(), product_namespace.to_owned());
```
Techassi: note: I would add the parameter at the previous step (where `install:manifests` is called). This would only require a clone of the product namespace instead of the complete hash map.

note: I would use `productNamespace` as the templating variable instead of `NAMESPACE` to be inline with other variables we already provide.

rust/stackablectl/CHANGELOG.md
```diff
@@ -4,11 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Pass the namespace as argument `NAMESPACE` to templated manifests.
+  This should unblock demos to run in all namespaces, as they can template the namespace e.g. for the fqdn of services ([#355]).
```
Techassi: suggestion: Adjust templating variable name

```suggestion
- Pass the product namespace as a templating variable `productNamespace` to manifests.
  This should unblock demos to run in all namespaces, as they can template the namespace e.g. for the FQDN of services ([#355]).
```

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -85,7 +90,7 @@ pub trait InstallManifestsExt {
                     })?;
 
                     let helm_chart: helm::Chart = transfer_client
-                        .get(&helm_file, &Template::new(parameters).then(Yaml::new()))
+                        .get(&helm_file, &Template::new(&parameters).then(Yaml::new()))
```
Techassi, line 93: note: If the above clone is removed, this and the borrow below can be removed again.

Techassi (changes requested):

rust/stackablectl/CHANGELOG.md
sbernauer: Yeah, I see points for productNamespace vs NAMESPACE.

productNamespace:
* Similar to stack arguments
* More aligned with how the CLI flag is called

NAMESPACE:
* Similar to our kuttl tests
* The different casing makes it clear that this is a special variable, and not coming from stack.yaml (other than the others)
* One could argue it's needless to prefix "product" namespace, as there is no other potential namespace here.

sbernauer: Picked the stylistic changes into 4905016

rust/stackable-cockpit/src/platform/manifests.rs
sbernauer: I intentionally added it to `install_manifests`, as it is called in `StackSpec::install_manifests` as well as `DemoSpec::install_manifests`.
This way we avoid duplicated code and logic.
I think the clone is acceptable im comparison to code complexity/maintainability.

Regarding the naming we have https://github.com/stackabletech/stackable-cockpit/pull/355#discussion_r1978813195, I will update the PR accordingly.

rust/stackablectl/CHANGELOG.md
Techassi: I'm still in favor of keeping it identical to the CLI argument. There might come a time where we also want the operator namespace to be available as a templating variable, and I would like make it very clear what namespace is being talked about in the template.

rust/stackable-cockpit/src/platform/manifests.rs
Techassi: Alright, let's keep the implementation as is. The naming will discussed in https://github.com/stackabletech/stackable-cockpit/pull/355#discussion_r1973171123.

rust/stackablectl/CHANGELOG.md
sbernauer: Ok, can we agree on `PRODUCT_NAMESPACE` than?

NickLarsenNZ: I disgree with using `PRODUCT_NAMESPACE`. Products are one thing, but the demos and stacks are more than the SDP products.

I also think `stackablectl` should use `--operator-namespace` and `--namespace` (well known), rather than `--operator-namespace` and `--product-namespace` (bespoke, and less applicable when it comes to demos/stacks).

See: https://github.com/stackabletech/stackable-cockpit/issues/371

sbernauer: I have to admin I'm with Nick here

Techassi: I don't have strong feelings for `PRODUCT_NAMESPACE` vs `NAMESPACE`. I'm more worried about consistency between the templating variable and the CLI argument. They must be the same.

I'm fine with getting this PR in and changing the CLI arguments in a follow-up PR. I just want to avoid us forgetting to change it.

I will only agree to a new release once the naming is the same in both places.

NickLarsenNZ: I agree with the consistency, and don't mind if it is done in a follow up PR.
I would vote for removing `product` from the arg since it is not only about products.

ie:
- `--namespace` (defaults to `default`)
- `--operator-namespace` (defaults to `stackable-operators`)

:bulb: We could keep the old `--product-namespace` argument around for a release and emit a warning to use the `--namespace` arg instead.

NickLarsenNZ: PR for the `--namespace` argument here: https://github.com/stackabletech/stackable-cockpit/pull/373

sbernauer: Thanks @NickLarsenNZ! The PR looks good to me

rust/stackable-cockpit/src/platform/manifests.rs
sbernauer: See https://github.com/stackabletech/stackable-cockpit/pull/355/files#r2007550029, not removing the clone

rust/stackablectl/CHANGELOG.md
sbernauer: Updated this PR, please have another look :)

NickLarsenNZ (approved):
LGTM

rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -97,18 +97,14 @@ impl DemoSpec {
...
+    pub async fn check_prerequisites(&self, client: &Client, namespace: &str) -> Result<(), Error> {
         debug!("Checking prerequisites before installing demo");
 
         // Returns an error if the demo doesn't support to be installed in the
         // requested namespace
-        if !self.supports_namespace(product_namespace) {
+        if !self.supports_namespace(namespace) {
             return Err(Error::UnsupportedNamespace {
-                requested: product_namespace.to_string(),
+                requested: namespace.to_string(),
```
Techassi: suggestion: Wile we are here, use `to_owned()` instead.

```suggestion
                requested: namespace.to_owned(),
```

rust/stackable-cockpit/src/platform/stack/spec.rs
```diff
@@ -111,21 +111,17 @@ impl StackSpec {
...
         // Returns an error if the stack doesn't support to be installed in the
         // requested product namespace. When installing a demo, this check is
         // already done on the demo spec level, however we still need to check
         // here, as stacks can be installed on their own.
-        if !self.supports_namespace(product_namespace) {
+        if !self.supports_namespace(namespace) {
             return Err(Error::UnsupportedNamespace {
                 supported: self.supported_namespaces.clone(),
-                requested: product_namespace.to_string(),
+                requested: namespace.to_string(),
```
Techassi: suggestion: Same as above.

```suggestion
                requested: namespace.to_owned(),
```

Techassi (dismissed):

Techassi:
One more note: Please make sure that the pre-commit checks pass.

NickLarsenNZ (dismissed):
LGTM

NickLarsenNZ (approved):
LGTM

PR #360, opened 2025-03-13
fix: Install correct operator version when using OCI registry 

Techassi:
When installing any operator using

```bash
stackablectl operator install ...
```

without specifying the version or chart source, this command installed the latest stable version (in this case, 24.11.1). Instead, it should install the `0.0.0-dev` version when no exact version using `=...` is specified.

The issue boils down to different handling of versions / version constraints. If the user didn't provide a specific version, stackablectl defaulted to `>0.0.0-0`, which worked for index.yaml based repositories but installed the wrong version for OCI registries.

stackablectl now uses `0.0.0-dev` as the default. Both chart sources now behave as expected.

### DIFF
--- a/rust/stackable-cockpit/src/constants.rs
+++ b/rust/stackable-cockpit/src/constants.rs
@@ -25,7 +25,7 @@ pub const HELM_REPO_INDEX_FILE: &str = "index.yaml";
 pub const HELM_OCI_BASE: &str = "oci.stackable.tech";
 pub const HELM_OCI_REGISTRY: &str = "oci://oci.stackable.tech/sdp-charts";
 
-pub const HELM_DEFAULT_CHART_VERSION: &str = ">0.0.0-0";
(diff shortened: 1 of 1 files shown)


adwk67 (commented):
This works for individual operators, but demos still install 24.11.1 by default. Is that intentional?  (it makes sense, as users will probably want the last stable version by default).

adwk67 (approved):
LGTM. Tested with single operators. Demos can be installed in a specific version with e.g. `--release dev` or `--release 24.11` (defaulting to the latest, 24.11)

Techassi:
> This works for individual operators, but demos still install 24.11.1 by default. Is that intentional? (it makes sense, as users will probably want the last stable version by default).

Yes, for now it is intentional. We want to change the default behaviour (to install latest stable operator by default) after the 25.3.0 release.

PR #369, opened 2025-04-10
chore: Bump kube, k8s_openapi and stackable-operator

Techassi:
This bumps a bunch of dependencies and applies the 2024 formatting. We cannot use the 2024 edition yet, because `cc` needs support for unsafe extern blocks.

This also gets rid of the `opentelemetry-jaeger` transitive dependency.


Cargo.toml
```diff
@@ -11,6 +11,8 @@ resolver = "2"
 version = "0.0.0-dev"
 authors = ["Stackable GmbH <info@stackable.tech>"]
 license = "OSL-3.0"
+# We can use 2024 as soon as https://github.com/rust-lang/cc-rs/issues/1409 is
+# fixed / supported.
```
NickLarsenNZ, line 15: praise!

Cargo.toml
```diff
@@ -46,7 +48,7 @@ serde_json = "1.0"
 serde_yaml = "0.9"
 sha2 = "0.10"
 snafu = { version = "0.8", features = ["futures"] }
-stackable-operator = { git = "https://github.com/stackabletech/operator-rs.git", tag = "stackable-operator-0.74.0" }
+stackable-operator = { git = "https://github.com/stackabletech/operator-rs.git", tag = "stackable-operator-0.91.1", default-features = false }
```
NickLarsenNZ: Just curious, what happens when we bump to 0.92.0?

My understanding was that dependencies can use whatever edition they want and it doesn't change the edition used here.

Techassi: Yes, I think that's how it should work. But I vaguely remember there being a different issue. Let me try it though.

Techassi: Or I just misremembered and it straight up works. See 7602e945c368a4c5db1c246271308120ae954aed.

PR #373, opened 2025-04-22
refactor: Rename `--product-namespace` to `--namespace`

NickLarsenNZ:
Rename `--product-namespace` to `--namespace` (the old argument still works and will be removed in a later release).

Also rename fields in various structs to make clear what they are for.


sbernauer (approved):
Thanks!

Techassi (approved):

Issue #153, opened 2023-12-01
feat: Implement progress reporting with third-party tool stdout support


Techassi:
Currently, there is no output until the command completes (either successfully or with an error).
Some operations take considerable time and leave the user wondering if anything is happening.

It is possible to set the log level, but that is too verbose for normal operation and is more for debugging.

```[tasklist]
### Tasks
- [ ] Print output as tasks are in progress or complete
- [ ] Add a progress spinner
- [ ] Capture STDOUT/STDERR from Command calls
```

PR #376, opened 2025-05-16
feat: add progress reporting

xeniape:
# Description

Implementation for https://github.com/stackabletech/stackable-cockpit/issues/153

Adds various progress indications/visualizations to `stackablectl` subcommand outputs.

demo install - operators

demo install - charts

**Some explanation regarding `println!`/`eprintln!` changes in this PR:**
When indicatif is active, it needs to control how text is printed. So instead of printing straight to `stdout`/`stderr` via `println!`/`eprintln!`, we use `indicatif_println!`/`indicatif_eprintln!` so it can deal with it while rendering progress bars. If it is not active, the indicatif macros just use `stdout`/`stderr`.

```[tasklist]
# Author
- [x] Changelog updated
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -189,6 +191,7 @@ impl DemoSpec {
         transfer_client: &xfer::Client,
     ) -> Result<(), Error> {
         info!("Installing demo manifests");
+        Span::current().pb_set_style(&ProgressStyle::with_template("{spinner} Installing manifests").unwrap());
```
NickLarsenNZ: We should use `expect` instead  of `unwrap` here, that way we can explain why the error case should never occur.

NickLarsenNZ: Same with all other occurrences or unwrap.

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -74,6 +76,11 @@ pub trait InstallManifestsExt {
     ) -> Result<(), Error> {
         debug!("Installing manifests");
 
+        Span::current().pb_set_style(
+            &ProgressStyle::with_template("Progress: {wide_bar} {pos}/{len}").unwrap(),
+        );
+        Span::current().pb_set_length(manifests.len() as u64);
+
```
NickLarsenNZ: These can be chained if you want:


```suggestion
        Span::current().pb_set_style(
            &ProgressStyle::with_template("Progress: {wide_bar} {pos}/{len}").expect("valid progress template"),
        )
        .pb_set_length(manifests.len() as u64);

```

rust/stackable-cockpit/src/platform/operator/mod.rs
```diff
@@ -221,7 +231,7 @@ impl OperatorSpec {
     {
         match helm::uninstall_release(&self.helm_name(), namespace.as_ref(), true) {
             Ok(status) => {
-                println!("{status}");
+                indicatif_println!("{}", status);
```
NickLarsenNZ: It would be worth explaining somewhere (PR description) why we use indicatif_println.

Something like:

> When indicatif is active, it needs to control how text is printed. So instead of printing straight to `stdout`/`stderr` via `println!`/`eprintln!`, we use `indicatif_println!`/`indicatif_eprintln!` so it can deal with it while rendering progress bars.
>
> If it is not active, the indicatif macros just use `stdout`/`stderr`.

rust/stackablectl/src/cmds/operator.rs
```diff
@@ -327,9 +337,8 @@ async fn install_cmd(args: &OperatorInstallArgs, cli: &Cli) -> Result<String, Cm
             )
             .context(HelmSnafu)?;
 
-        // TODO (@NickLarsenNZ): Send this to the progress handler instead of using println
-        // info!(%operator, "Installed operator");
-        println!("Installed {} operator", operator);
+        info!(%operator, "Installed operator");
+        indicatif_println!("Installed {} operator", operator);
```
NickLarsenNZ: We should try to avoid unnamed interpolations:

```suggestion
        indicatif_println!("Installed {operator} operator");
```

rust/stackablectl/src/main.rs
```diff
@@ -16,21 +18,45 @@ async fn main() -> Result<(), Error> {
...
+                    .pretty()
+                    .with_writer(indicatif_layer.get_stderr_writer()),
+            )
+            .init();
+    } else {
+        tracing_subscriber::registry()
+            .with(level_filter)
+            .with(indicatif_layer)
+            .init();
+    }
```
NickLarsenNZ: I think this should be the case no matter which level was set. But I am guessing clap has a default for when the user hasn't provided it (rather than using an Option).

We would need to revisit this.

We could either remove the if/else, or change the clap arg to an Option so that we can check if it was set.

rust/stackablectl/src/main.rs
```diff
@@ -16,21 +18,45 @@ async fn main() -> Result<(), Error> {
...
 
     // Load env vars from optional .env file
     match dotenv() {
         Ok(_) => (),
         Err(err) => {
             if !err.not_found() {
-                println!("{err}")
+                indicatif_println!("{err}")
```
NickLarsenNZ: Even though it didn't before, we might want to emit this to stderr:


```suggestion
                indicatif_eprintln!("{err}")
```

NickLarsenNZ (commented):
Really nice. Here's some initial feedback.

I noticed a lot of formatting changes, but I couldn't tell if it was new or old styling.
I also see pre-commit failed, so could be related.

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: Unfortunately, those can't be chained, the `pb_*` functions don't have a return value

rust/stackable-cockpit/src/platform/operator/mod.rs
xeniape: Added the explanation! Thank you very much :)

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: https://github.com/stackabletech/stackable-cockpit/pull/376/commits/6d80e341512eca4abfc65bc610b112ae5f6fac58

rust/stackablectl/src/cmds/operator.rs
xeniape: https://github.com/stackabletech/stackable-cockpit/pull/376/commits/6d80e341512eca4abfc65bc610b112ae5f6fac58

rust/stackablectl/src/main.rs
xeniape: https://github.com/stackabletech/stackable-cockpit/pull/376/commits/6d80e341512eca4abfc65bc610b112ae5f6fac58

rust/stackable-cockpit/src/platform/manifests.rs
NickLarsenNZ: No problem.

rust/stackablectl/src/main.rs
xeniape: https://github.com/stackabletech/stackable-cockpit/pull/376/commits/d03f5482e528b960c27762a94246771725c4d297

rust/stackable-cockpit/src/platform/operator/mod.rs
```diff
@@ -186,6 +194,8 @@ impl OperatorSpec {
         chart_source: &ChartSourceType,
     ) -> Result<(), helm::Error> {
         info!(operator = %self, "Installing operator");
+        Span::current().pb_set_message(format!("Installing {}-operator", self.name).as_str());
```
NickLarsenNZ:
```suggestion
        Span::current().pb_set_message(format!("Installing {name}-operator", name = self.name).as_str());
```

rust/stackable-cockpit/src/platform/operator/mod.rs
```diff
@@ -61,10 +63,15 @@ pub struct OperatorSpec {
...
-            None => "".into(),
-        })
+        write!(
+            f,
+            "{}{}",
+            self.name,
+            match &self.version {
+                Some(v) => format!("={v}"),
+                None => "".into(),
+            }
```
NickLarsenNZ:
```suggestion
            "{name}{version_selector}",
            name = self.name,
            version_selector = match &self.version {
                Some(v) => format!("={v}"),
                None => "".into(),
            }
```

NickLarsenNZ (approved):
LGTM, just a couple of optional changes

rust/stackablectl/src/cmds/stack.rs
```diff
@@ -303,14 +306,15 @@ fn describe_cmd(
     }
 }
 
-#[instrument(skip(cli, stack_list, transfer_client))]
+#[instrument(skip(cli, stack_list, transfer_client), fields(indicatif.pb_show = true))]
```
NickLarsenNZ, line 309: Note to self, we should `skip_all` here, since args ends up looking so bad in the logs (since it uses the Debug impl). Nothing related to this PR though.

NickLarsenNZ (approved):
LGTM

PR #382, opened 2025-06-02
chore: Release stackablectl 1.0.0

Techassi:
This releases stackablectl **1.0.0**.

This is the **first** version released with the new versioning in place. Previously, the tool used the same versioning as our SDP releases, which lead to confusions, especially around patch-level releases. This is why we decided to use an independent versioning mechanism going forward.

### Added

- Pass the stack/demo namespace as a templating variable `NAMESPACE` to manifests.
  This should unblock demos to run in all namespaces, as they can template the namespace e.g. for the FQDN of services ([#355]).
- Add progress reporting ([#376]).
- Include SparkConnectServer in the `stacklet list` output ([#380]).
- Add release upgrade functionality to `stackablectl release` command through `upgrade` subcommand ([#379]).

### Changed

- Renamed `--product-namespace` argument to `--namespace` ([#373], [#355]).
  - Kept `--product-namespace` as a hidden alias to be removed in a later release.

### Fixed

- Prefix `ui-http` port endpoints with `http://`, as e.g. used by hbase-operator ([#368]).

[#355]: https://github.com/stackabletech/stackable-cockpit/pull/355
[#368]: https://github.com/stackabletech/stackable-cockpit/pull/368
[#373]: https://github.com/stackabletech/stackable-cockpit/pull/373
[#376]: https://github.com/stackabletech/stackable-cockpit/pull/376
[#380]: https://github.com/stackabletech/stackable-cockpit/pull/380
[#379]: https://github.com/stackabletech/stackable-cockpit/pull/379

PR #383, opened 2025-06-02
ci: Bump Ubuntu runners

Techassi:
Part of #382.

During the release workflow, a job was cancelled because it used an outdated Ubuntu runner (20.04). All Ubuntu runners now use 24.04 instead.

### DIFF
--- a/.github/workflows/release_stackablectl.yml
+++ b/.github/workflows/release_stackablectl.yml
@@ -24,10 +24,10 @@ jobs:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-20.04 # We don't use ubuntu-latest because we want to build against an old glibc version. (18.04 has glibc 2.27, 20.04 has glibc 2.31, 22.04 has glibc 2.35)
+            os: ubuntu-24.04
             file-suffix: ""
           - target: aarch64-unknown-linux-gnu
-            os: ubuntu-22.04-arm # 2025-01: The ARM runners only support Ubuntu 22.04 or 24.04, so we can't pick 20.04:
+            os: ubuntu-24.04-arm
             file-suffix: ""
           - target: x86_64-apple-darwin
             os: macos-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release_stackablectl.yml
```diff
@@ -24,10 +24,10 @@ jobs:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-20.04 # We don't use ubuntu-latest because we want to build against an old glibc version. (18.04 has glibc 2.27, 20.04 has glibc 2.31, 22.04 has glibc 2.35)
```
sbernauer, line 27: Can we please only bump to 22.04 and keep the comment on why we want to?
(same below)

Techassi: But we don't need to? It builds perfectly fine with newer Ubuntu (and glibc) versions.

sbernauer: Ok, the comment is missing information I have to admit.

I don't have the time to dig into it, but AFAIK it's fine for the binary to run on newer glibc than we build against. But it can not run on a older version as it was build with.
To support as many linux distros (and versions) out there we try to compile against the oldest glibc we can find.

See https://github.com/stackabletech/stackablectl/pull/235

It would be awesome if you could extend the comment by this information

Techassi:
> LGTM, assuming there is no longer a glibc workaround needed

As far as I can see there is no need to use an older glibc version anymore. The binaries build fine in CI and on my local machine (which is also based on Ubuntu 24.04).

Merging as is, we can still rollback if needed.

.github/workflows/release_stackablectl.yml
Techassi: Marking as resolved, see https://github.com/stackabletech/stackable-cockpit/pull/383#issuecomment-2931005376.

sbernauer: Created a follow-up PR: https://github.com/stackabletech/stackable-cockpit/pull/407

Issue #187, opened 2022-08-08
Research option to uninstall Stacks and Demos


sbernauer:
**Update after a discussion on 2025-11-26:**

In case, a user wants to try out a demo and uninstall it afterwards, it should be documented, how to do this. stackablectl should provide an option for it.

- All resources with a Stackable label should be removed, including the data (PersistentVolumes). That means that all top-level resources, which are not (indirectly) owned by a Stackable labeled resource, must be labeled accordingly, e.g. the namespace a demo is installed in.
- Demos and stacks should be installed in the non-default namespace.
- External tools like PostgreSQL in the demo/stacks namespace will therefore also be removed.
- The first version of this removal "tool" shows a warning and provides no option to be less destructive and keep resources.

---
**Initial description:**

As a user 
I want to be able to uninstall a stack or demo by itself, after I'm done looking at it
to have a clean cluster again and not have to remove things manually, which is error prone.

- What should be uninstalled, should operators be left behind?
- What about multiple stacks in the same namespace? What about shared things like a ZooKeeper in two stacks. we _could_ try and have only one ZK and label it as being part of two stacks, in a sort of dependency tree.
- Should a demo be always in a dedicated namespace? Deleting a namespace will not delete cluster-scoped resources
- What about removing Redis/Postgres etc. that is part of a demo?
- What about data? I.e. PVCs? maybe there could be `uninstall` vs `purge` and `purge` also removes data.
- What about cluster scoped resources? How do we make sure they do not comflict with each other?

Removing

- some Manifests need to be deleted
- Helm stuff needs to be uninstalled (i.e. Postgres)

Related ADR: https://docs.stackable.tech/home/stable/contributor/adr/adr031-resource-labels

Random unrelated code dump, do not execute :P
```bash
kubectl delete ns stackable-operators
kubectl get crd -o name | grep stackable | xargs kubectl delete
kubectl delete --all pvc
```

fhennig:
I just updated the ticket a bit to include some new thoughts we had and also tried to incorporate the discussion in here so far.

lfrancke:
I'd like to have a more detailed refinemend first with concrete steps that we can achieve independently from each other.

> All resources with a Stackable label should be removed, including the data (PersistentVolumes). That means that all top-level resources, which are not (indirectly) owned by a Stackable labeled resource, must be labeled accordingly, e.g. the namespace a demo is installed in.

It's unclear to me how much work is behind this.
[...]

PR #429, opened 2026-03-12
feat: Add uninstall demo/stack feature

xeniape:
# Description

Part of https://github.com/stackabletech/stackable-cockpit/issues/187

This PR adds the `uninstall` subcommands to `demo` and `stack` commands. It implements the MVP features mentioned here https://github.com/stackabletech/stackable-cockpit/issues/187#issuecomment-3748248917

The flag of skipping operators and CRD deletion was optional and I wasn't sure about the naming or if needs to be one command or two. Can also be just removed, added it because it was not much additional work.

Tested with all current demos/stacks. Some demos have edge cases in the deletion, will document that in the parent issue.

~This PR also contains the `DEMO`/`STACK` parameter addition for https://github.com/stackabletech/demos/pull/374#discussion_r2916588046~ stripped out into https://github.com/stackabletech/stackable-cockpit/pull/432

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


docs/modules/stackablectl/pages/commands/demo.adoc
```diff
@@ -159,4 +159,31 @@ $ stackablectl stacklets list
...
+
+[source,console]
+----
+$ stackablectl demo uninstall trino-taxi-data -n <NAMESPACE>
+
+----
+
+[NOTE]
+====
+The uninstall command deletes the namespace the demo was installed in. Therefore it is not possible to uninstall demos in the `default` namespace.
```
NickLarsenNZ, line 172: I feel like this should be opt-in behaviour. Or rather, if we created the namespace, it gets deleted.

There could be existing namespaces that shouldn't be deleted.

rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -180,33 +191,138 @@ impl DemoSpec {
...
+    ))]
+    pub async fn uninstall(
+        &self,
+        release_list: ReleaseList,
+        uninstall_parameters: DemoUninstallParameters,
+        client: &Client,
+        transfer_client: &xfer::Client,
+        stack: StackSpec,
+    ) -> Result<(), Error> {
+        // Uninstall Helm Charts
```
Techassi, line 214: note: I think this comments seems misplaced and needs to move a few lines lower.

docs/modules/stackablectl/pages/commands/demo.adoc
NickLarsenNZ: One such example is if we have the monitoring stack installed, and want to add and remove demos.

docs/modules/stackablectl/pages/commands/stack.adoc
```diff
@@ -89,3 +89,39 @@ Installed stack logging
...
+
+[source,console]
+----
+$ stackablectl stack uninstall logging -n <NAMESPACE>
+
+----
+
+[NOTE]
+====
+The uninstall command deletes the namespace the stack was installed in. Therefore it is not possible to uninstall stacks in the `default` namespace.
```
NickLarsenNZ, line 105: Same as my comment about demos.

rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -180,33 +191,138 @@ impl DemoSpec {
...
+                &uninstall_parameters.demo_namespace,
+                &ApiResource::from_gvk(&GroupVersionKind {
+                    group: "".to_owned(),
+                    version: "v1".to_owned(),
+                    kind: "Namespace".to_owned(),
+                }),
+                None,
+            )
+            .await
+            .context(DeleteObjectSnafu)?;
```
Techassi, line 253: note: I think we should abstract that away in an associated function called `delete_namespace` similar to what we have to create a namespace. See:

https://github.com/stackabletech/stackable-cockpit/blob/32cf8475a73b2daff38c3af48bcff3f7c9ac141b/rust/stackable-cockpit/src/utils/k8s/client.rs#L443-L460

NickLarsenNZ, line 253: Here's where I think we should have a check to see if the namespace was "owned" by the demo (via labels/annotations).

rust/stackable-cockpit/src/platform/demo/spec.rs
```diff
@@ -180,33 +191,138 @@ impl DemoSpec {
...
+
+        // Delete remaining objects not namespace scoped
+        client
+            .delete_all_objects_with_label(
+                "stackable.tech/demo",
+                &uninstall_parameters.demo_name,
```
Techassi: question: Do we want to use a validated `Label` instead of a free-form string here?

docs/modules/stackablectl/pages/commands/demo.adoc
xeniape: According to https://github.com/stackabletech/stackable-cockpit/issues/187#issue-2112087523 "Update after a discussion on 2025-11-26:" it was decided the first version provides no option to be less destructive. Not sure which meeting that was or if it makes sense to change the requirements now.

rust/stackable-cockpit/src/platform/stack/spec.rs
```diff
@@ -64,14 +69,20 @@ pub enum Error {
...
+    DeleteObject { source: k8s::Error },
 }
 
 /// This struct describes a stack with the v2 spec
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 #[cfg_attr(feature = "openapi", derive(ToSchema))]
 pub struct StackSpec {
-    /// A short description of the demo
+    /// A short description of the stack
```
Techassi, line 88: Hehe, nice catch!

docs/modules/stackablectl/pages/commands/demo.adoc
xeniape: There were also problems with deleting everything without deleting the namespace, see https://github.com/stackabletech/hdfs-operator/pull/761#issuecomment-4000035927. So maybe for the first version / MVP it is fine for now like this? But feel free to disagree

rust/stackable-cockpit/src/platform/stack/spec.rs
```diff
@@ -197,12 +204,96 @@ impl StackSpec {
...
+                &uninstall_parameters.stack_namespace,
+                &ApiResource::from_gvk(&GroupVersionKind {
+                    group: "".to_owned(),
+                    version: "v1".to_owned(),
+                    kind: "Namespace".to_owned(),
+                }),
+                None,
+            )
+            .await
+            .context(DeleteObjectSnafu)?;
```
NickLarsenNZ, line 245: and here

rust/stackable-cockpit/src/platform/demo/spec.rs
NickLarsenNZ: One thing we could consider is reading in the demo, get the GVKs and Helm charts, and uninstall in reverse order before uninstalling the namespace (if it was created by the demo).

Same would apply for stacks.

That could be a follow-up PR, but I think at least we should be careful with namespace deletion.

NickLarsenNZ (commented):
Left a couple of comments about namespace deletion.

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -85,31 +91,15 @@ pub trait InstallManifestsExt {
         // We need some additional templating capabilities, e.g. the namespace, so that stacks/demos
         // can use that to render e.g. the fqdn service names [which contain the namespace].
         parameters.insert("NAMESPACE".to_owned(), namespace.to_owned());
-        parameters.insert("STACK".to_owned(), stack_name.into());
-        if let Some(demo_name) = demo_name {
-            parameters.insert("DEMO".to_owned(), demo_name.into());
-        }
```
Techassi, line 91: question: Just that we are on the same track: These are now added earlier and are part of the passed `parameters`, right?

question: Out of curiosity, what was the reason you did that? I assume the demo and stack name are needed earlier in the call chain now?

docs/modules/stackablectl/pages/commands/demo.adoc
NickLarsenNZ: I left an idea below in how we could do it: https://github.com/stackabletech/stackable-cockpit/pull/429#discussion_r2986721449

But of course, if this implements what we agreed on, that's fine.
It does feel a little dangerous though.

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -168,4 +158,65 @@ pub trait InstallManifestsExt {
...
+                        helm_chart.version, "Uninstalling Helm chart",
+                    );
+
+                    helm::uninstall_release(&helm_chart.release_name, namespace, true).context(
+                        UninstallHelmReleaseSnafu {
+                            release_name: &helm_chart.release_name,
+                        },
+                    )?;
+                }
+                ManifestSpec::PlainYaml(_) => {}
```
Techassi: question: Why are we only uninstalling Helm manifests and not all?

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -168,4 +158,65 @@ pub trait InstallManifestsExt {
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+pub async fn get_helmchart(
```
Techassi: note: We should use a consistent spelling, either `helm_chart` or `helmchart`. I prefer the first one.

rust/stackable-cockpit/src/utils/k8s/client.rs
```diff
@@ -287,6 +319,116 @@ impl Client {
...
+        let api_resources = self
+            .get_api_resources()
+            .await
+            .into_iter()
+            .filter(|(_, capability)| {
+                capability.supports_operation(discovery::verbs::LIST)
+                    && match namespace {
+                        Some(_) => capability.scope == kube::discovery::Scope::Namespaced,
+                        None => capability.scope == kube::discovery::Scope::Cluster,
+                    }
```
Techassi: note: I think this is a little hard to read/grasp. I would try to split this boolean expression into two `let` assignments before ANDing them.

rust/stackable-cockpit/src/utils/k8s/client.rs
```diff
@@ -287,6 +319,116 @@ impl Client {
...
+                                &object.metadata.name.unwrap(),
+                                &api_resource,
+                                object.metadata.namespace.as_deref(),
+                            )
+                            .await?;
+                        }
+                    }
+                }
+            }
+        }
```
Techassi: note: I think we should try reducing the amount of right-shift here. We might be able to use something like:

```rust
let Some(object_list) = objects else {
    continue;
}
```

rust/stackable-cockpit/src/utils/k8s/client.rs
```diff
@@ -287,6 +319,116 @@ impl Client {
...
+                    delete_request = object_api
+                        .delete(object_name, &DeleteParams::foreground())
+                        .await;
+                }
+                // Right side of Either, when deletion is done
+                either::Either::Right(_) => {
+                    return Ok(());
+                }
+            }
+        }
```
Techassi: note: We might want to use kube's `is_deleted` for this instead: https://docs.rs/kube/latest/kube/runtime/wait/conditions/fn.is_deleted.html

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -382,6 +410,33 @@ async fn install_cmd(
...
+                    Ok(args.demo_name.clone())
+                } else {
+                    // User selected to install in default namespace
+                    Ok(args.namespaces.namespace.clone())
+                }
+            } else {
+                Ok(args.namespaces.namespace.clone())
+            }
+        },
+    )?;
```
Techassi: note: Well that's a mouthful to read. Can we somehow make this easier to grasp?

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -382,6 +410,33 @@ async fn install_cmd(
...
+    // `stackablectl demo uninstall` relies on namespace deletion, suggest installing in a non-default namespace
+    // It should still be possible to skip that if either uninstall is not needed
+    // or installing an older version of the demo which only supports the 'default' namespace
+    let demo_namespace = tracing_indicatif::suspend_tracing_indicatif(
+        || -> Result<String, CmdError> {
+            if args.namespaces.namespace == DEFAULT_NAMESPACE {
+                if Confirm::new()
+                .with_prompt(
+                    format!(
+                        "Demos installed in the '{DEFAULT_NAMESPACE}' namespace cannot be uninstalled with stackablectl. Install the demo in the '{demo_namespace}' namespace instead?",
```
Techassi: note: Use the debug impl for `String` instead, which already includes quotes:


```suggestion
                        "Demos installed in the {DEFAULT_NAMESPACE:?} namespace cannot be uninstalled with stackablectl. Install the demo in the {demo_namespace:?} namespace instead?",
```

docs/modules/stackablectl/pages/commands/demo.adoc
xeniape: Yes, I agree it feels dangerous, tried to avoid it at first but ran into the problems above, which unfortunately won't be solved by your idea I think. The manifests we would delete with that approach were also deleted with the help of the labels, so would be the same situation, unless I misunderstood something :disappointed: 

Another thing might be adding a confirmation dialogue like the one I added to the installation for the namespace. That it asks if the user really wants to uninstall since that would delete the namespace, and cancel the deletion otherwise.

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -447,3 +499,70 @@ async fn install_cmd(
...
+
+    let stack_files = cli
+        .get_stack_files(release_branch)
+        .context(PathOrUrlParseSnafu)?;
+    let stack_list = stack::StackList::build(&stack_files, transfer_client)
+        .await
+        .context(BuildListSnafu)?;
+
+    // Get the stack spec based on the name defined in the demo spec
+    let stack = stack_list.get(&demo.stack).context(NoSuchStackSnafu {
```
Techassi: note: already extracting the stack here (and then having to clone it via `to_owned` below) is a completely different approach compared to installing a demo. Why don't we use the same approach here as well?

Techassi (commented):
Left a few comments and questions.

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: Those were more meant like segmentation comments about the various steps. The code statements below all belongs to that Helm uninstall part. I don't think the `uninstall_helm_manifests` function needs the comment saying basically the same as the function name. Can just remove it, if it's more confusing than helping out.

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: Sounds good, thanks! [`e71437a` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/e71437a43e81f089a83c97e6210ea3b071dc82fe)

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: [`4d1ff73` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/4d1ff73659e054c07a72e7780d86d224f774c1fe)

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: I didn't add them in `install_manifests` because it seemed like this function is "outside of the scope of demos and stacks" and might be also used for something else. Along with the comment above that templating shouldn't be a concern here in the future. Since @sbernauer pulled that parameter addition from this PR while I was gone, I don't know what his thoughts were, we didn't coordinate. But I'm fine with also moving it back into `install_manifests`.

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: Because those would already be covered by the namespace deletion and the deletion by labels. Used the ADR as reference also https://docs.stackable.tech/home/stable/contributor/adr/adr031-resource-labels/#_stackablectl_demo_uninstall_name

Should I add a comment there, or would you say the approach in itself should be different?

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: [`6d0edef` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/6d0edefb548f5b286c2aba0442e8f810930574a9)

rust/stackable-cockpit/src/utils/k8s/client.rs
xeniape: [`2367a88` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/2367a889cf8f79fe5c9a9b49a02dc7367f32c741) hope this improves it!

rust/stackable-cockpit/src/utils/k8s/client.rs
xeniape: [`1a00526` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/1a005268ea043781551ce4a0f78fb1f0ce96744a) thanks!

rust/stackable-cockpit/src/utils/k8s/client.rs
xeniape: Ohhh, found a new function also thanks to that suggestion, this is a huge improvement now thanks :smile: [`043f0f1` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/043f0f19b569d8cec9353092c759cd718cd89767)

Only tested it with two demos, but it seemed to behave the same way as the previous implementation

rust/stackablectl/src/cmds/demo.rs
xeniape: Maybe I looked at the code too long at this point and unable to see it, but could you give me more information on what is making it hard to grasp. At least I currently don't know how to make it shorter. Or are there explanations missing somewhere? Or just the whole "formatting" of it?

rust/stackablectl/src/cmds/demo.rs
xeniape: Hmm not sure, might've been an oversight [`1bc08a2` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/1bc08a251a21f9a5b8e4b7faa470d0e5179a74ea)

docs/modules/stackablectl/pages/commands/demo.adoc
xeniape: [`0f3cf7f` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/0f3cf7f557d651f1d9b1d9f6ee365e6da7ff4499) Maybe this addresses your concern a bit, let me know what you think :)

rust/stackablectl/src/cmds/demo.rs
xeniape: [`0f3cf7f` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/0f3cf7f557d651f1d9b1d9f6ee365e6da7ff4499) Tried rearranging it a bit, let me know if know if that makes it any better

rust/stackable-cockpit/src/platform/demo/spec.rs
Techassi: Ah I see, then let's leave it as is.

rust/stackable-cockpit/src/platform/demo/spec.rs
Techassi: Any particular reason why we are not using `Api<Namespace>` instead? I guess one upside to to doing it like currently implemented is that we don't need to clone the client for a single function call.

rust/stackable-cockpit/src/utils/k8s/client.rs
```diff
@@ -287,6 +320,86 @@ impl Client {
...
+                )
+                .await?;
+
+            let Some(object_list) = objects else {
+                continue;
+            };
+
+            for object in object_list {
+                if let Some(value) = object.labels().get(&label.key().to_string()) {
+                    if value.eq(&label.value().to_string()) {
```
Techassi: note: We could potentially use `deref` here to save an allocation.


```suggestion
                    if value.eq(label.value().deref()) {
```

rust/stackable-cockpit/src/utils/k8s/client.rs
```diff
@@ -287,6 +320,86 @@ impl Client {
...
+                    namespace,
+                )
+                .await?;
+
+            let Some(object_list) = objects else {
+                continue;
+            };
+
+            for object in object_list {
+                if let Some(value) = object.labels().get(&label.key().to_string()) {
```
Techassi: note: This is kinda cumbersome to use, but could be improved by https://github.com/stackabletech/operator-rs/pull/1182 (if we decide to move forward with it).

rust/stackable-cockpit/src/platform/manifests.rs
Techassi: Ah yeah, didn't even remember this ADR. I think the current approach is fine, but slightly harder to wrap your head around, because we delete different things in different places.

If we would handle deletions of manifests here, the "source of deletion" is more focused (at least that is my current impression). We could then finalize the deletion by deleting the namespace.

If we decide to keep it as is, a comment should definitely be added here.

rust/stackable-cockpit/src/utils/k8s/client.rs
Techassi: What do you think about renaming `scope_matching` to `capability_scope_matches`?

rust/stackable-cockpit/src/utils/k8s/client.rs
Techassi: Huh, neat! I also didn't know _that_ function exists.

Nice change!

Techassi (commented):
Only a few small comments left, mostly looks nice!

rust/stackablectl/src/cmds/demo.rs
Techassi: I don't feel like it does make it better, sorry :see_no_evil: 

What might help on the other hand is to define the closure outside of the `tracing_indicatif::suspend_tracing_indicatif` call.

If that also doesn't help, we can leave it as is.

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: The client is cloned anyway in the `delete_object` function called by `delete_namespace`. I think I didn't use `Api<Namespace>` was because I wanted to reuse the `delete_object` function (before everything was moved into `delete_namespace`). Could possibly also rewrite that one to use `Api<DynamicObject>` instead of `ApiResource`. I don't remember exactly anymore, but I think the resulting code just looked better using `ApiResource` and that's what the discovery already provided. But no strong opinions here.

While writing, tried implementing it since then the Api creation could be moved from `delete_object` to `delete_all_object_with_label`, so one client clone per object list, instead of object function call. This doc for `DynamicObject`:

> /// A dynamic representation of a kubernetes object
> ///
> /// This will work with any non-list type object.

made me think, that if a function accepts `Api<DynamicObject>`, I could pass it a `Api<Namespace>` (or any other Object type). But seems to be not compatible. So I think I'm abandoning that implementation again.

Could also not use `delete_object` instead, but then I would just re-implement what's in that function anyway.

Just noting down thoughts here while trying it out, not trying to argue against it. Here the non-working diff because of `Api<DynamicObject>` <-> `Api<Namespace>` incompatibility.

```
diff --git a/rust/stackable-cockpit/src/utils/k8s/client.rs b/rust/stackable-cockpit/src/utils/k8s/client.rs
index 79f5338..779ec7f 100644
--- a/rust/stackable-cockpit/src/utils/k8s/client.rs
+++ b/rust/stackable-cockpit/src/utils/k8s/client.rs
@@ -358,13 +358,20 @@ impl Client {
                 continue;
             };
 
+            let object_api = match namespace {
+                Some(namespace) => {
+                    Api::<DynamicObject>::namespaced_with(self.client.clone(), namespace, &api_resource)
+                }
+                None => Api::<DynamicObject>::all_with(self.client.clone(), &api_resource),
+            };
+
             for object in object_list {
                 if let Some(value) = object.labels().get(&label.key().to_string()) {
                     if value.eq(&label.value().to_string()) {
                         self.delete_object(
                             &object.metadata.name.unwrap(),
-                            &api_resource,
-                            object.metadata.namespace.as_deref(),
+                            &api_resource.kind,
+                            object_api.clone(),
                         )
                         .await?;
                     }
@@ -379,17 +386,10 @@ impl Client {
     pub async fn delete_object(
         &self,
         object_name: &str,
-        api_resource: &ApiResource,
-        namespace: Option<&str>,
+        object_kind: &str,
+        object_api: Api<DynamicObject>,
     ) -> Result<(), Error> {
-        Span::current().pb_set_message(&format!("Deleting {} {}", api_resource.kind, object_name));
-
-        let object_api = match namespace {
-            Some(namespace) => {
-                Api::<DynamicObject>::namespaced_with(self.client.clone(), namespace, api_resource)
-            }
-            None => Api::<DynamicObject>::all_with(self.client.clone(), api_resource),
-        };
+        Span::current().pb_set_message(&format!("Deleting {} {}", object_kind, object_name));
 
         kube::runtime::wait::delete::delete_and_finalize(
             object_api,
@@ -574,14 +574,11 @@ impl Client {
 
     /// Deletes a [`Namespace`] with `name` in the cluster.
     pub async fn delete_namespace(&self, name: String) -> Result<()> {
+        let namespace_api: Api<Namespace> = Api::all(self.client.clone());
         self.delete_object(
             &name,
-            &ApiResource::from_gvk(&GroupVersionKind {
-                group: "".to_owned(),
-                version: "v1".to_owned(),
-                kind: "Namespace".to_owned(),
-            }),
-            None,
+            "Namespace",
+            namespace_api,
         )
         .await
     }

```

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: Talked with @sbernauer, he is fine wherever that code resides, as long as the `STACK` and `DEMO` parameters are provided for both demos and stacks (if installed through a demo). That logic shouldn't have been broken by my change I hope.

rust/stackable-cockpit/src/platform/manifests.rs
xeniape: [`506fbcf` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/506fbcf8446671fb8121e91d2f60c1d152097480) Left a comment for now, not sure about the formulation though, since it didn't feel right adding anything that implies knowledge about the process using this function. Hope this one is a good compromise

rust/stackable-cockpit/src/utils/k8s/client.rs
xeniape: Yeah sure [`a6e2feb` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/a6e2febe025c480e1040ec3ea48b3a1ef5c57bba) :)

rust/stackablectl/src/cmds/demo.rs
xeniape: [`18e7f70` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/18e7f70d9b70c11c265327734d96a274acc5594a)

rust/stackable-cockpit/src/utils/k8s/client.rs
Techassi: The linked PR was merged btw, but not yet released.

rust/stackablectl/src/cmds/demo.rs
Techassi: I already like this way better. I think we can now also switch to the previous style of assigning `demo_namespace` which should make it even cleaner.

rust/stackable-cockpit/src/platform/manifests.rs
Techassi: Alright cool, marking as resolved.

rust/stackable-cockpit/src/platform/demo/spec.rs
Techassi: > The client is cloned anyway in the `delete_object` function called by `delete_namespace`.

Ah I see. Then I think we should use the same style as in `create_namespace` via `Api`.

rust/stackablectl/src/cmds/demo.rs
xeniape: [`cdbd185` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/cdbd185edd8a21b3270847cfa706a975870d617c) did you mean like this?

rust/stackable-cockpit/src/platform/demo/spec.rs
xeniape: [`30f6d96` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/30f6d96e943b15ac9cb25d07c1c1749e3ad447b9)

rust/stackablectl/src/cmds/demo.rs
Techassi: Yup, exactly! Nicely done.

Techassi (dismissed):

rust/stackable-cockpit/src/utils/k8s/client.rs
xeniape: [`55eeceb` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/55eeceb5fbe4079a3c8aa6014a7cb8171bcfa084)

rust/stackablectl/src/cmds/demo.rs
```diff
@@ -447,3 +498,89 @@ async fn install_cmd(
...
+                format!(
+                    "Uninstalling the demo {demo_name:?} will delete the {demo_namespace:?} namespace. This action cannot be undone. Proceed?",
+                demo_name = args.demo_name.clone(),
+                demo_namespace = args.namespaces.namespace.clone())
+            )
+            .default(true)
+            .interact()
+            .context(ConfirmDialogSnafu)
+        },
+    )?;
```
NickLarsenNZ: Curious what happens when this runs in a script? Does indicative choose something, or just hang?

docs/modules/stackablectl/pages/commands/demo.adoc
NickLarsenNZ: This is fine. I left another comment about what happens if used in a script.
https://github.com/stackabletech/stackable-cockpit/pull/429#discussion_r3050416163

Marking this as resolved.

rust/stackable-cockpit/src/platform/demo/spec.rs
NickLarsenNZ: Will mark as resolved (we won't do it now. Let's wait to see how we do it in BBQ - maybe something reusable there).

rust/stackablectl/src/cmds/demo.rs
xeniape: Tried it out, this would result in an error:

```
An unrecoverable error occurred: failed to execute demo (sub)command

 1: failed to confirm user input
 2: IO error: No such device or address (os error 6)
```

There is the `.interact_opt()` function, which theoretically might return `None` if it's not a terminal the command is running in. But it would also return a `None` if the user presses `Esc`/`q` to cancel, but those two should behave differently, so can't distinguish based on that.

I think your suggestion of having a `--yes`/`-y` option would be the way to go here.

xeniape: Tried it out here [`14f9f7b` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/14f9f7b8601bfa7a03519511c7e139b101074db1) (and also fixed some smaller things I found through that). Feedback welcome :)

rust/stackable-cockpit/src/utils/k8s/client.rs
Techassi: Neat!

rust/stackablectl/CHANGELOG.md
```diff
@@ -6,8 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Heuristic for setting listener presets when the environment is Minikube ([#438]).
 - Add `uninstall` subcommand for `demo`/`stack` commands ([#429]).
+- Add confirmation prompt to `install` subcommand for namespace selection ([#429]).
+- Add `--assume-yes` option for running commands non-interactively ([#429]).
 
+[#438]: https://github.com/stackabletech/stackable-cockpit/pull/438
```
Techassi: note: That line should go below this PR (429).

Techassi (dismissed):

rust/stackablectl/CHANGELOG.md
xeniape: [`d468c04` (this PR)](https://github.com/stackabletech/stackable-cockpit/pull/429/commits/d468c0407e260783d34f9d62e89bbdda026e2978)

Techassi (approved):

PR #432, opened 2026-03-17
feat: Add `STACK` and `DEMO` templating parameters

sbernauer:
# Description

Part of https://github.com/stackabletech/stackable-cockpit/issues/187

Extracted the `STACK` and `DEMO` templating out of https://github.com/stackabletech/stackable-cockpit/pull/429, as it's the IMHO uncontroversial part. It unblocks us from adapting the demo YAMLs already (https://github.com/stackabletech/demos/pull/374).
Doing so I fixed some problems regarding what to make available when, it now has the same semantics as the existing stack and demo labels. I also document them in the README.

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -65,11 +65,13 @@ pub enum Error {
 pub trait InstallManifestsExt {
     // TODO (Techassi): This step shouldn't care about templating the manifests nor fetching them from remote
     #[instrument(skip_all, fields(%namespace, indicatif.pb_show = true))]
```
NickLarsenNZ, line 67: Maybe we want to add the new parameters as fields.
Doesn't have to be now, just an idea.

Note: Option needs special treatment to look nice, but I'd have to dig up an example from existing code.

NickLarsenNZ (dismissed):
Nice! LGTM

rust/stackable-cockpit/src/platform/manifests.rs
sbernauer: Currently, we call this fn at two places, and AFAIK don't plan to call it at more.
My brain already exploded by so many structs basically representing the same properties but being slightly different.
As it's ok for you I would prefer to skip that for now.

NickLarsenNZ (approved):

PR #437, opened 2026-03-31
chore: Add script to verify release notes

Techassi:
This PR adds a script to verify that a release notes partial exists for the current stackablectl version specified via Cargo.toml.


NickLarsenNZ (approved):
Nice!

PR #440, opened 2026-04-22
feat: stacks and demo supports OCI helmCharts

### LINKED ISSUE #417: Allow stacks/demos to deploy helmChart from OCI registry
It seems to not be possible to deploy a helm chart sourced from an OCI registry.

Here's what I tried (using the Stackable nifi-operator as an example):

A demo (I assume the same is true for a stack)

### `demos-v2.yaml`

```yaml
  some-demo:
    stackableStack: some-stack
    manifests:
      - helmChart: ./path/to/nifi-operator.yaml
    # other required fields...
```

### `nifi-operator.yaml`

```yaml
releaseName: nifi-operator
name: nifi-operator
repo:
  name: nifi-operator
  url: oci://oci.stackable.tech/sdp-charts/nifi-operator
version: 0.0.0-dev
options: {} # override values
```

Error when applying:

```
An unrecoverable error occured: demo command error

 1: failed to install demo "some-demo"
 2: failed to install stack manifests
 3: failed to install Helm release nifi-operator
 4: failed to upgrade/install Helm release
 5: helm FFI library call failed (no cached repo found. (try 'helm repo update'): open /tmp/.helmcache/nifi-operator-index.yaml: no such file or directory)
```

NickLarsenNZ:
Fixes #417

> [!NOTE]
> This was bugging me enough when trying to validate the observability stack (which before this requires a bunch of manual steps).

You can now refer to Helm Charts stored in OCI.

Eg: in a stack/demo yaml:

```yaml
- helmChart: release-file.yaml
```

`release-file.yaml`:

```yaml
releaseName: nifi-operator
name: nifi-operator
repo:
  name: nifi-operator
  url: oci://oci.stackable.tech/sdp-charts # 👈 yeah boi
version: 0.0.0-dev
options: {}
```

Here's an example:

image

I also tested to make sure existing stacks still work (they do):

```sh
cargo stackablectl stack install logging
```

### DIFF
--- a/rust/stackable-cockpit/src/helm.rs
+++ b/rust/stackable-cockpit/src/helm.rs
@@ -542,50 +542,24 @@ where
 
 #[cfg(test)]
 mod tests {
+    use rstest::rstest;
+
     use super::*;
 
-    #[test]
-    fn source_kind_oci() {
+    #[rstest]
+    #[case("oci://oci.stackable.tech/sdp-charts", ChartSourceKind::Oci)]
+    #[case(
+        "https://repo.stackable.tech/repository/helm-stable",
+        ChartSourceKind::Repo
+    )]
+    #[case("http://example.com/charts", ChartSourceKind::Repo)]
+    #[case("./charts/my-chart", ChartSourceKind::Local)]
+    #[case("/absolute/path/to/chart", ChartSourceKind::Local)]
+    fn source_kind(#[case] url: &str, #[case] expected: ChartSourceKind) {
         let repo = ChartRepo {
-            name: "nifi-operator".to_string(),
-            url: "oci://oci.stackable.tech/sdp-charts".to_string(),
+            name: "test".to_owned(),
+            url: url.to_owned(),
         };
-        assert_eq!(repo.source_kind(), ChartSourceKind::Oci);
-    }
-
-    #[test]
-    fn source_kind_https_repo() {
-        let repo = ChartRepo {
-            name: "stackable-stable".to_string(),
-            url: "https://repo.stackable.tech/repository/helm-stable".to_string(),
-        };
-        assert_eq!(repo.source_kind(), ChartSourceKind::Repo);
-    }
-
-    #[test]
-    fn source_kind_http_repo() {
-        let repo = ChartRepo {
-            name: "example".to_string(),
-            url: "http://example.com/charts".to_string(),
-        };
-        assert_eq!(repo.source_kind(), ChartSourceKind::Repo);
-    }
-
-    #[test]
-    fn source_kind_relative_path_is_local() {
-        let repo = ChartRepo {
-            name: "local".to_string(),
-            url: "./charts/my-chart".to_string(),
-        };
-        assert_eq!(repo.source_kind(), ChartSourceKind::Local);
-    }
-
-    #[test]
-    fn source_kind_absolute_path_is_local() {
-        let repo = ChartRepo {
-            name: "local".to_string(),
-            url: "/absolute/path/to/chart".to_string(),
-        };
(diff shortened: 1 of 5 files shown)


rust/stackable-cockpit/src/helm.rs
```diff
@@ -505,3 +539,53 @@ where
 
     serde_yaml::from_str(&index_file_content).context(DeserializeYamlSnafu)
 }
+
+#[cfg(test)]
+mod tests {
```
Techassi, line 544: note: We could use rstest for these (if it already is a dependency)

rust/stackable-cockpit/src/helm.rs
```diff
@@ -505,3 +539,53 @@ where
...
+    use super::*;
+
+    #[test]
+    fn source_kind_oci() {
+        let repo = ChartRepo {
+            name: "nifi-operator".to_string(),
```
Techassi: note: We should use `to_owned` instead.

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -56,6 +56,10 @@ pub enum Error {
         source: helm::Error,
     },
 
+    /// This error indicates that the Helm chart source kind is not supported.
+    #[snafu(display("local Helm chart sources are not yet supported (url: {url})"))]
+    UnsupportedChartSource { url: String },
```
Techassi: question: Do we still want to call this `url`, even tho it could be a local path? Maybe `source` would be a better name, but I don't know how that interacts with snafu.

rust/stackable-cockpit/src/platform/manifests.rs
```diff
@@ -56,6 +56,10 @@ pub enum Error {
         source: helm::Error,
     },
 
+    /// This error indicates that the Helm chart source kind is not supported.
+    #[snafu(display("local Helm chart sources are not yet supported (url: {url})"))]
```
Techassi: note: For strings, we use the Debug impl for printing.

```suggestion
    #[snafu(display("local Helm chart sources are not yet supported (url: {url:?})"))]
```

Techassi (commented):
LGTM, left a few small comments.

rust/stackable-cockpit/src/helm.rs
NickLarsenNZ: Ahh man. I even know that

rust/stackable-cockpit/src/platform/manifests.rs
NickLarsenNZ: Good point. This wasn't initially dealing with local paths until I wondered what other "strings" a person could put in. Will fix.

NickLarsenNZ: Ahh `source` conflicts with Snafu. Will make it similar.

rust/stackable-cockpit/src/platform/manifests.rs
NickLarsenNZ: Done in b3d6cd7

rust/stackable-cockpit/src/platform/manifests.rs
NickLarsenNZ: Done in b3d6cd7

rust/stackable-cockpit/src/helm.rs
NickLarsenNZ: Done in c519e4b

rust/stackable-cockpit/src/helm.rs
NickLarsenNZ: It was, so done in c519e4b

Techassi (approved):

PR #450, opened 2026-08-17
ci: Harden workflows

Saul-STFC:
# Description

This PR adds CI Hardening to the stackablectl repo 

This addresses all warnings from Zizmor and also bumps dependency versions.

NOTE: - `sigstore/cosign-installer` v3.9.1 → v4.1.2 :  v3+ of Cosign has breaking changes to the default signature format, i'm not sure how that affects our registry/signing setup, so `cosign-release` is explicitly pinned to `v2.6.5` for now rather than taking the installer's new default.

See: 
https://github.com/sigstore/cosign/releases/tag/v3.0.1
https://blog.sigstore.dev/cosign-3-0-available/
https://github.com/goharbor/harbor/issues/22592

### Author

- [ ] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


Saul-STFC:
@NickLarsenNZ @Techassi

.github/workflows/pr_cockpit.yml
```diff
@@ -90,22 +95,26 @@ jobs:
...
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Install cosign
-        uses: sigstore/cosign-installer@398d4b0eeef1380460a10c8013a76f728fb906ac # v3.9.1
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+        with:
+          # pin to the latest Cosign v2.x, V3 changes how we publish (new bundle format / OCI 1.1 referrer signatures)
+          cosign-release: v2.6.5
```
Saul-STFC: https://github.com/stackabletech/stack_scanner/blob/master/.github/workflows/scan_release.yml

Stackscanner uses the newer version so this doesn't seem to be needed?

NickLarsenNZ: I'll ask @dervoeti

dervoeti: Yes, the pinning should be removed, for docker-images and operators this is already the case.

Saul-STFC: https://github.com/stackabletech/csaf_publisher/blob/main/.github/workflows/build_docker_image.yml
https://github.com/stackabletech/actions/blob/386bc87db059dcba817e968eb96db773648e549f/publish-image-index-manifest/action.yaml#L51

I imagine it's ok as other repos use it

.github/workflows/pr_cockpit.yml
Techassi, line 1: I would be fine removing this entire workflow as we just use the underlying library but don't support/use the container image + helm chart anymore.

.github/workflows/pr_docs.yml
Techassi, line 1: This can also go.

.github/workflows/pr_general.yml
Techassi, line 1: This should also go and all checks/lints should be handled by pre-commit/prek.

.github/workflows/release_stackablectl.yml
```diff
@@ -1,10 +1,11 @@
 ---
 name: Stackablectl Release Pipeline
+permissions: {}
 
 on:
   push:
     tags:
-      - "stackablectl-[0-9]+.[0-9]+.[0-9]+**"
+      - "stackablectl-[0-9]+.[0-9]+.[0-9]+*"
```
Techassi: The `**` is different compared to `*`. The change should technically be fine, but was there a particular reason you changed this?

.github/workflows/pr_pre-commit.yml
```diff
@@ -13,10 +14,12 @@ env:
 
 jobs:
   pre-commit:
+    name: Pre-commit
```
Techassi:
```suggestion
    name: pre-commit
```

.github/workflows/pr_prek.yaml
Techassi, line 1: I think we should switch to `prek` instead.

.github/workflows/pr_stackablectl.yml
```diff
@@ -75,16 +80,18 @@ jobs:
...
+        run: cargo build --target "$TARGET" -p stackablectl
 
       - name: Ensure shell completions up-to-date
-        if: matrix.os == 'x86_64-unknown-linux-gnu'
+        if: matrix.target == 'x86_64-unknown-linux-gnu'
         run: |
           cargo xtask gen-comp
           git diff --exit-code
 
       - name: Ensure man page is up-to-date
```
Techassi: Those two checks should instead be handled by pre-commit/prek.

.github/workflows/release_stackablectl.yml
```diff
@@ -38,34 +41,42 @@ jobs:
...
+        run: mv "target/$TARGET/release/stackablectl$FILE_SUFFIX" "stackablectl-$TARGET$FILE_SUFFIX"
 
       - name: Rename SBOM
-        run: mv rust/stackablectl/stackablectl_bin.cdx.xml stackablectl-${{ matrix.target }}.cdx.xml
+        env:
+          TARGET: ${{ matrix.target }}
+        run: mv rust/stackablectl/stackablectl_bin.cdx.xml "stackablectl-$TARGET.cdx.xml"
 
       - name: Upload Release Binary and SBOM
         uses: softprops/action-gh-release@72f2c25fcb47643c292f7107632f7a47c1df5cd8 # v2.3.2
```
Techassi: I think we should move to a `gh release ...` approach, similar to other reworks. See https://github.com/stackabletech/docker-images/blob/main/.github/workflows/boil_release.yaml for example.

Techassi (changes requested):
The changes themself are fine, but I think we should use the opportunity to improve the workflows in general. That is something which is on my plate since forever, but I just didn't get around to it yet.

.github/workflows/pr_cockpit.yml
Saul-STFC: Done, #792de01a

.github/workflows/pr_docs.yml
Saul-STFC: Done, #792de01a

.github/workflows/pr_general.yml
Saul-STFC: Done, #792de01a

.github/workflows/release_stackablectl.yml
Saul-STFC: This wasn't meant to be changed and I have reverted it, I think I must have fat-fingered a vim motion

.github/workflows/pr_prek.yaml
Saul-STFC: Done, #792de01a

.github/workflows/pr_stackablectl.yml
Saul-STFC: Done, #792de01a

.github/workflows/release_stackablectl.yml
Saul-STFC: Done, #792de01a

Techassi (approved):

